PR #1, opened 2018-09-17
[KIECLOUD-3] Baseline operator functionality

bmozaffa:
[KIECLOUD-3] Baseline operator functionality that sets up a hardcoded RHPAM trial environment


sutaakar:
@bmozaffa It looks like the commit contains also third party classes. Wouldn't it be possible to import them somehow from external location instead (by Maven or any other tool)?

bmozaffa:
@sutaakar It is and they are imported using tooling. I'm not an expert in Go, but from everything I read, the recommended practice is to check in your vendor directory. There is apparently just no stable central repo like maven, and libraries can disappear, which is why they suggest doing this.

sutaakar:
@bmozaffa Ok, sounds reasonable.

PR #13, opened 2018-10-16
[KIECLOUD-22] keystore & secret generation; go & sdk upgrade

tchughesiv:
[KIECLOUD-22] keystore & secret generation; go & sdk upgrade

 - keystore, password, and secret object generated during deployment
 - go version upgraded to v1.10.4 (required for marshaling key to pkcs8 encoded form)
 - operator-sdk upgraded to v0.0.7 (requires k8s v1.11.2)


config/app/envs/prod.yaml
```diff
@@ -301,7 +287,9 @@ servers:
       name: view
 
 ## RANGE BEGINS
-{{$Name := .ApplicationName}}{{ range $index, $Num := .ServerCount }}
+{{$Name := .ApplicationName}}
+{{$KeyStorePW := .KeyStorePassword}}
+{{ range $index, $Num := .ServerCount }}
```
bmozaffa: Do you really need to introduce a new/local variable for each loop, even though you are using the same global var in the loop?

internal/app/handler/handler.go
```diff
@@ -88,10 +91,41 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 
 func NewEnv(cr *opv1.App) ([]runtime.Object, error) {
 	var objs []runtime.Object
-	env, err := defaults.GetEnvironment(cr)
+	env, pw, err := defaults.GetEnvironment(cr)
```
bmozaffa: Please use "password", or if you can't, passwd or similar that is more clear and less ambiguous.

internal/pkg/defaults/defaults.go
```diff
@@ -9,36 +9,42 @@ import (
...
-	env := v1.Environment{}
+func GetEnvironment(cr *opv1.App) (v1.Environment, []byte, error) {
+	var env v1.Environment
 	// default to '1' Kie Server
 	if cr.Spec.NumKieServers == 0 {
 		cr.Spec.NumKieServers = 1
 	}
 
+	// pw := []byte("mykeystorepass")
+	pw := shared.GeneratePassword(8)
```
bmozaffa: Same here: Please use "password", or if you can't, passwd or similar that is more clear and less ambiguous.

internal/pkg/shared/utils.go
```diff
@@ -37,6 +48,111 @@ func getEnvVars(defaults map[string]string, vars []corev1.EnvVar) []corev1.EnvVa
 	return allVars
 }
 
+func GenerateKeystore(alias string, pw []byte) []byte {
```
bmozaffa: Same here: please use "password", or if you can't, passwd or similar that is more clear and less ambiguous.

internal/pkg/shared/utils.go
```diff
@@ -37,6 +48,111 @@ func getEnvVars(defaults map[string]string, vars []corev1.EnvVar) []corev1.EnvVa
...
+	var b bytes.Buffer
+	err = keystore.Encode(&b, keyStore, pw)
+	if err != nil {
+		logrus.Error(err)
+	}
+
+	return b.Bytes()
+}
+
+func genCert() (cert []byte, derPK []byte, err error) {
```
bmozaffa: Ruben is doing something similar in the APB, please sync up with him, would be good to get confirmation that we're not missing anything and vice-versa.

config/app/envs/prod.yaml
tchughesiv: unfortunately, based on what i've seen.  I guess i could try to duplicate the same vars within the ServerCount slice instead... have each var duplicated within the EnvTemplate struct, hande it there.

bmozaffa: Ok, that's fine, was just wondering why you were doing that.

internal/pkg/shared/utils.go
tchughesiv: i've made improvements, most notably creating the https routes before generating certs so that the operator can use the cluster-assigned route host/url as its CN.

PR #15, opened 2018-11-07
[KIECLOUD-27] set/override env vars from cr ; authoring envs added

tchughesiv:
[KIECLOUD-27]
 - improved env var add/override capability from CR
 - consolidated common container configs to new yaml file (only env vars for now)
 - added ability to set Rhpam version and imagetag from CR
 - new envs added -
   - authoring-ha
   - authoring


sutaakar:
@tchughesiv Just for clarification, is it possible to configure those versions and imagetag by user? If so, how?

tchughesiv:
@sutaakar sure, the user would just set the required version/tag in the CR yaml before creating it in OpenShift. e.g. 
```yaml
apiVersion: kiegroup.org/v1
kind: App
metadata:
  name: authoring
spec:
  environment: authoring
  version: "7.0"
  imageTag: "1.2"
  kieDeployments: 2
  console:
    resources:
      limits:
        memory: 2Gi
    env:
    - name: key1
      value: value1
  server:
    resources:
      limits:
        memory: 2Gi
```

config/app/commonConfigs.yaml
```diff
@@ -0,0 +1,46 @@
+objects:
+  console:
+    env:
+    - name: KIE_ADMIN_USER
+      value: adminUser
+    - name: KIE_ADMIN_PWD
+      value: "{{.AppPassword}}"
```
bmozaffa: These 3 passwords, when not specified, should all be generated to different values. Of course the same passwords should be reused in the server component for them to work together.

However, in the trial environment, when not specified, they should all be hardcoded to "RedHat"

internal/pkg/defaults/defaults.go
```diff
@@ -6,38 +6,64 @@ import (
...
+		cr.Spec.KieDeployments = 1
+	}
+	if cr.Spec.Version == "" {
+		cr.Spec.Version = constants.RhpamVersion
+	}
+	if cr.Spec.ImageTag == "" {
+		cr.Spec.ImageTag = constants.ImageStreamTag
 	}
 
 	//password := []byte("mykeystorepass")
```
bmozaffa: Remove this?

internal/pkg/defaults/defaults.go
```diff
@@ -6,38 +6,64 @@ import (
...
+	commonBytes, err := loadYaml("commonConfigs.yaml", envTemplate)
+	if err != nil {
+		return env, nil, err
+	}
+	var common opv1.AppSpec
+	err = yaml.Unmarshal(commonBytes, &common)
+	if err != nil {
+		logrus.Error(err)
+	}
+	// envs.env <- common.env <- cr.env
```
bmozaffa: Remove?

config/app/commonConfigs.yaml
tchughesiv: done... i hard-code the trial passwords within the CR, as its designed to do.

internal/pkg/defaults/defaults.go
tchughesiv: done

internal/pkg/defaults/defaults.go
tchughesiv: done

deploy/cr-trial.yaml
```diff
@@ -4,20 +4,42 @@ metadata:
...
+      - name: KIE_SERVER_CONTROLLER_USER
+        value: controllerUser
+      - name: KIE_SERVER_CONTROLLER_PWD
+        value: RedHat
+      - name: KIE_SERVER_USER
+        value: executionUser
+      - name: KIE_SERVER_PWD
+        value: RedHat
+      - name: RHPAMCENTR_MAVEN_REPO_PASSWORD
+        value: RedHat
```
bmozaffa: Should be in the env file, not the CR. The CR should be empty to get the default environment behavior, and would provide passwords only to override what's the default. In fact, we shouldn't really distribute more than one CR as an example, but that change can wait till later.

tchughesiv: done

config/app/envs/trial.yaml
```diff
@@ -22,36 +22,29 @@ console:
...
+              requests:
+                memory: 768Mi
+              limits:
+                memory: 1Gi
             livenessProbe:
               exec:
                 command:
                 - /bin/bash
                 - -c
-                - curl --fail --silent -u adminUser:RedHat http://localhost:8080/kie-wb.jsp
```
bmozaffa: hmm, does the user setting their own password cause this probe to fail?

tchughesiv: It does cause it to fail, which is why I changed the probes to leverage the env vars directly instead.

PR #17, opened 2018-11-26
[KIECLOUD-37] add ImageStreams locally if don't exist in openshift ns

tchughesiv:
[KIECLOUD-37]
Checks DeploymentConfigs for image change triggers/tags.
Creates local ImageStreamTag if one doesn't already exist in the `openshift` or local namespaces.

 - imageClient added to reconciler
 - openshift authorization packages no longer needed... switched to k8s rbac
 - improved Makefile logic


pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -346,15 +401,28 @@ func (reconciler *ReconcileKieApp) createCustomObjects(object v1.CustomObject, c
...
 	for _, obj := range object.DeploymentConfigs {
 		obj.SetGroupVersionKind(oappsv1.SchemeGroupVersion.WithKind("DeploymentConfig"))
 		controllerutil.SetControllerReference(cr, &obj, reconciler.scheme)
 		obj.SetNamespace(cr.Namespace)
+		for ti, trigger := range obj.Spec.Triggers {
+			if trigger.Type == oappsv1.DeploymentTriggerOnImageChange {
+				if !reconciler.checkImageStreamTag(trigger.ImageChangeParams.From.Name, trigger.ImageChangeParams.From.Namespace) {
+					err := reconciler.createLocalImageTag(trigger.ImageChangeParams.From, cr)
+					if err != nil {
+						logrus.Error(err)
```
bmozaffa: What now? Should it bubble up an error and fail to deploy?

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -37,7 +40,11 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileKieApp{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	imageClient, err := imagev1.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		logrus.Error(err)
```
bmozaffa: Again, doesn't an error mean we can't move forward?

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -185,30 +193,72 @@ func (reconciler *ReconcileKieApp) Reconcile(request reconcile.Request) (reconci
...
-		return reconcile.Result{}, err
+		return false
 	}
-	return reconcile.Result{Requeue: true}, nil
+	return true
+}
+
+// Create ImageStreamTag
+func (reconciler *ReconcileKieApp) createLocalImageTag(currentTagReference corev1.ObjectReference, cr *v1.KieApp) error {
+	if !reconciler.checkImageStreamTag(currentTagReference.Name, cr.Namespace) {
```
bmozaffa: Isn't this a duplicate check? Seems you called this function because IS did not exist. Also, the function should be meaningful on its own, and its job is to create the local image tag, not to create it "if not exists"... right?

pkg/controller/kieapp/kieapp_controller.go
tchughesiv: i decided they DC should still deploy even if the imagestreamtag doesn't succeed. not sure what the benefit is of stopping the env deployment.

pkg/controller/kieapp/kieapp_controller.go
tchughesiv: i decided they DC should still deploy even if the imagestreamtag doesn't succeed. not sure what the benefit is of stopping the env deployment.

bmozaffa: Not sure... is it 100% that the DC will fail? If it is, I suppose the benefit is user experience, in that instead of going through DC deployment errors and being confused as to why it's not working, they would see the IS error and focus on fixing it.

pkg/controller/kieapp/kieapp_controller.go
bmozaffa: Sorry, just seeing this now! My only concern was what we see much too often: a bunch of errors that are all red herrings, because the real errors was ignored earlier on. I guess that's not the case here, right? Because everything else will be successful, though the pods won't start, and the only error will be the one you log here?

tchughesiv: ah... maybe a warn instead? idk... ill think about other options.

pkg/controller/kieapp/kieapp_controller.go
tchughesiv: DC won't "fail" ... it'll create fine, just won't trigger deployment until imagestream exists. On the plus side, openshift UI will give clues as to why the deploy hasn't started.

PR #19, opened 2018-11-28
[KIECLOUD-47] First few steps towards a comprehensive merge framework

bmozaffa:
The provided merge.go is not 100% complete, but goes a long way towards merging OpenShift objects. The production-lite.yaml starts validating this approach by taking out just a small subset of its content and moving them to server.yaml, with merge_test.go validating the result.

The defaults.go class has undergone minimal refactor to allow the use of envTemplate logic without breaking current usage.

### DIFF
--- a/config/common/server.yaml
+++ b/config/common/server.yaml
@@ -1,0 +1,267 @@
+{{ range $index, $Map := .ServerCount }}
+persistentVolumeClaims:
+  - metadata:
+      name: "{{.ApplicationName}}-postgresql-claim-{{$index}}"
+      labels:
+        app: "{{.ApplicationName}}"
+        service: "{{.ApplicationName}}-postgresql-{{$index}}"
+    spec:
+      accessModes:
+        - ReadWriteOnce
+      resources:
+        requests:
+          storage: "1Gi"
+## PostgreSQL persistent volume claim END
+## KIE server deployment config BEGIN
+deploymentConfigs:
+  - metadata:
+      name: "{{.ApplicationName}}-kieserver-{{$index}}"
+      labels:
+        app: "{{.ApplicationName}}"
+    spec:
+      strategy:
+        type: Recreate
+      triggers:
+        - type: ImageChange
+          imageChangeParams:
+            automatic: true
+            containerNames:
+              - "{{.ApplicationName}}-kieserver"
+            from:
+              kind: ImageStreamTag
+              namespace: "openshift"
+              name: "rhpam{{.Version}}-kieserver-openshift:{{.ImageTag}}"
+        - type: ConfigChange
+      replicas: 3
+      selector:
+        deploymentConfig: "{{.ApplicationName}}-kieserver-{{$index}}"
+      template:
+        metadata:
+          name: "{{.ApplicationName}}-kieserverblh-{{$index}}"
+          labels:
+            deploymentConfig: "{{.ApplicationName}}-kieserver-{{$index}}"
+            app: "{{.ApplicationName}}"
+            service: "{{.ApplicationName}}-kieserver-{{$index}}"
+        spec:
+          serviceAccountName: "{{.ApplicationName}}-rhpamsvc"
+          terminationGracePeriodSeconds: 60
+          containers:
+            - name: "{{.ApplicationName}}-kieserver"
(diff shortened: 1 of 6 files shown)


bmozaffa:
@tchughesiv Please review!

config/common/server.yaml
```diff
@@ -0,0 +1,267 @@
...
+              kind: ImageStreamTag
+              namespace: "openshift"
+              name: "rhpam{{.Version}}-kieserver-openshift:{{.ImageTag}}"
+        - type: ConfigChange
+      replicas: 3
+      selector:
+        deploymentConfig: "{{.ApplicationName}}-kieserver-{{$index}}"
+      template:
+        metadata:
+          name: "{{.ApplicationName}}-kieserverblh-{{$index}}"
```
tchughesiv: appears there's some extra characters in this name? `blh` ?

config/common/server.yaml
```diff
@@ -0,0 +1,267 @@
+{{ range $index, $Map := .ServerCount }}
+persistentVolumeClaims:
```
tchughesiv: not sure if this can (or should) be included as a "common" kieserver component. Only prod uses postgres. Authoring, for example, uses a singular mysql DB.

config/common/server.yaml
```diff
@@ -0,0 +1,267 @@
...
+          imageChangeParams:
+            automatic: true
+            containerNames:
+              - "{{.ApplicationName}}-kieserver"
+            from:
+              kind: ImageStreamTag
+              namespace: "openshift"
+              name: "rhpam{{.Version}}-kieserver-openshift:{{.ImageTag}}"
+        - type: ConfigChange
+      replicas: 3
```
tchughesiv: replicas of 3 probably shouldn't be the default "common" setting. prod is the only env that uses this many.

config/common/server.yaml
```diff
@@ -0,0 +1,267 @@
...
+                value: executionUser
+              - name: KIE_SERVER_PWD
+                value: "{{.ServerPassword}}"
+              - name: DROOLS_SERVER_FILTER_CLASSES
+                value: "true"
+              - name: RHPAMCENTR_MAVEN_REPO_PATH
+                value: /maven2/
+              - name: KIE_SERVER_CONTROLLER_SERVICE
+                value: "{{.ApplicationName}}-rhpamcentr"
+              - name: KIE_MBEANS
```
tchughesiv: i believe there are a few env vars that aren't common here... including this MBEANS.

config/common/server.yaml
```diff
@@ -0,0 +1,267 @@
...
+              - name: SSO_OPENIDCONNECT_DEPLOYMENTS
+                value: ROOT.war
+              - name: KIE_SERVER_BYPASS_AUTH_USER
+                value: "false"
+              - name: KIE_SERVER_CONTROLLER_PROTOCOL
+                value: ws
+          volumes:
+            - name: kieserver-keystore-volume
+              secret:
+                secretName: "{{.ApplicationName}}-kieserver-{{$index}}-app-secret"
```
tchughesiv: this volume isn't common

tchughesiv:
@bmozaffa thinking about this more, maybe we should use the most simplistic env to build our first "common" files? trial? 
the more complex envs can then use the merge capability to add to it...

tchughesiv:
@bmozaffa all of that being said, I'm ok w/ merging as-is since the merge logic doesn't yet affect the operator's functionality. the default refactor LGTM

config/common/server.yaml
bmozaffa: This was actually meant to prove that the name is overwritten

config/common/server.yaml
bmozaffa: hmm, meant to set this to 1, must have forgotten.

bmozaffa:
@tchughesiv Cool. yeah the actual yaml files were only there as examples and to allow a working test file. Beyond that, I thought of it as a next step to come up with a proper baseline that is then overwritten.

I agree with server.yaml being a more common baseline, though that having been said, if a component is included in 4 templates and not a 5th, I would prefer to still have it in server.yaml and then delete it using the introduced annotation (delete = "true") in the fifth.

PR #20, opened 2018-11-30
Adding authoring-lite

ruromero:

### DIFF
--- a/config/common/server.yaml
+++ b/config/common/server.yaml
@@ -3,8 +3,9 @@ persistentVolumeClaims:
   - metadata:
       name: "{{.ApplicationName}}-postgresql-claim-{{$index}}"
       labels:
-        app: "{{.ApplicationName}}"
-        service: "{{.ApplicationName}}-postgresql-{{$index}}"
+        app: "{{.ApplicationName}}-postgresql-{{$index}}"
+        application: "{{.ApplicationName}}"
+        service: "{{.ApplicationName}}-postgresql"
     spec:
       accessModes:
         - ReadWriteOnce
@@ -17,7 +18,11 @@ deploymentConfigs:
   - metadata:
       name: "{{.ApplicationName}}-kieserver-{{$index}}"
       labels:
-        app: "{{.ApplicationName}}"
+        app: "{{.ApplicationName}}-kieserver-{{$index}}"
+        application: "{{.ApplicationName}}"
+        service: "{{.ApplicationName}}-kieserver"
+      annotations:
+        template.alpha.openshift.io/wait-for-ready: "true"
     spec:
       strategy:
         type: Recreate
@@ -37,11 +42,12 @@ deploymentConfigs:
         deploymentConfig: "{{.ApplicationName}}-kieserver-{{$index}}"
       template:
         metadata:
-          name: "{{.ApplicationName}}-kieserverblh-{{$index}}"
+          name: "{{.ApplicationName}}-kieserver-{{$index}}"
           labels:
+            app: "{{.ApplicationName}}-kieserver-{{$index}}"
+            application: "{{.ApplicationName}}"
+            service: "{{.ApplicationName}}-kieserver"
             deploymentConfig: "{{.ApplicationName}}-kieserver-{{$index}}"
-            app: "{{.ApplicationName}}"
-            service: "{{.ApplicationName}}-kieserver-{{$index}}"
         spec:
           serviceAccountName: "{{.ApplicationName}}-rhpamsvc"
           terminationGracePeriodSeconds: 60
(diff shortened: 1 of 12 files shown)


config/common/server.yaml
```diff
@@ -17,7 +18,11 @@ deploymentConfigs:
   - metadata:
       name: "{{.ApplicationName}}-kieserver-{{$index}}"
       labels:
-        app: "{{.ApplicationName}}"
+        app: "{{.ApplicationName}}-kieserver-{{$index}}"
```
tchughesiv: won't this split the kieservers up into separate groups in the UI?

bmozaffa: Agree, I think this is a mistake.

config/common/server.yaml
```diff
@@ -126,8 +166,9 @@ deploymentConfigs:
   - metadata:
       name: "{{.ApplicationName}}-postgresql-{{$index}}"
       labels:
-        app: "{{.ApplicationName}}"
-        service: "{{.ApplicationName}}-postgresql-{{$index}}"
+        app: "{{.ApplicationName}}-postgresql-{{$index}}"
+        application: "{{.ApplicationName}}"
+        service: "{{.ApplicationName}}-postgresql"
```
tchughesiv: not all kieservers will use postgres. i actually think only prod env uses it.

pkg/controller/kieapp/shared/utils.go
```diff
@@ -179,14 +179,11 @@ func envVarEqual(env corev1.EnvVar, envList []corev1.EnvVar) bool {
...
 	for _, cre := range src {
-		for envIndex, e := range dst {
-			if isEnvVarInList(cre.Name, dst) {
-				if cre.Name == e.Name {
-					dst[envIndex] = cre
-				}
-			} else {
-				dst = append(dst, cre)
-			}
+		pos, _ := getEnvVar(cre.Name, dst)
```
tchughesiv: does this change how the function works? was this a fix of some kind?

config/common/server.yaml
ruromero: Yes, as Babak said , I’m grouping them by using different labels but this obe is a mistake

config/common/server.yaml
ruromero: The authoring.yml file was deploying postgres, that’s why I made the kieserver use it

pkg/controller/kieapp/shared/utils.go
ruromero: It’s just avoiding an extra iteration. The outcome is the same

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -131,42 +132,42 @@ func getRoleBindingReferenceSlice(objects []rbacv1.RoleBinding) []v1.OpenShiftOb
...
-		return slice
 	}
+	slice := make([]appsv1.DeploymentConfig, combinedSize(baselineRefs, overwriteRefs))
+	err := mergeObjects(baselineRefs, overwriteRefs, slice)
+	if err != nil {
+		logrus.Errorf("%v", err)
+		return nil
+	}
+	return slice
+
```
bmozaffa: I keep looking at this, which looks like 100+ lines of diff, but I think is just a small cosmetic change to prevent an extra indent. Please revert so it's clear what's changed and what hasn't. We can have a pure refactor commit later on and include this.

config/common/server.yaml
tchughesiv: OK ... @bmozaffa are we ok w/ that? I'm not sure authoring-ha uses it. Does prod-immutable?

bmozaffa: Ok, so we have 5 templates... if 3 or 4 of them use something, we should make it common and delete it in the others. Otherwise, not make it common.

@ruromero I don't see authoring use Postgres in the latest template (under CE repo). It uses H2 there, and it's only prod/immutable, so I'd say to skip.

ruromero: @bmozaffa you're right. I don't know where I took the postgres deployment from. My bad. I'll go back to h2 or maybe make a copy for the moment just to not lose the postgres config (it took me a while)

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -131,42 +132,42 @@ func getRoleBindingReferenceSlice(objects []rbacv1.RoleBinding) []v1.OpenShiftOb
...
 		return baseline
-	} else if len(baseline) == 0 {
+	}
+	if len(baseline) == 0 {
 		return overwrite
-	} else {
-		baselineRefs := getDeploymentConfigReferenceSlice(baseline)
-		overwriteRefs := getDeploymentConfigReferenceSlice(overwrite)
-		for overwriteIndex := range overwrite {
-			overwriteItem := overwrite[overwriteIndex]
```
bmozaffa: @ruromero This is actually the issue that I previously tried to explain to you, that caused a bug elsewhere. Removing it introduces the same bug in this block. Your overwriteItem, which is the second return value of the range loop, is a copy and not the actual array element. Changes to it are not reflected in the overwrite array.

This introduces a second issue, which is that by trying to prove this with a test case, I realized much (if not all) of this code block is unnecessary. Mergo clearly has restrictions for what we are trying to do, but its restrictions are not as wide as I had assumed. It appears for example, that the mergo.merge of line 375 traverses a few levels deep into at least the content of ObjectMeta. Can you validate, and remove unnecessary code?

pkg/controller/kieapp/shared/utils.go
```diff
@@ -155,13 +155,13 @@ func GeneratePassword(length int) []byte {
 	return buf
 }
 
-func isEnvVarInList(envName string, Env []corev1.EnvVar) bool {
-	for _, v := range Env {
+func getEnvVar(envName string, Env []corev1.EnvVar) (int, *corev1.EnvVar) {
+	for pos, v := range Env {
 		if v.Name == envName {
-			return true
+			return pos, &v
```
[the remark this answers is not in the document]
ruromero: I don't want to return a copy. I prefer to return the pointer in case this value has to be changed which couldn't be achieved by returning a copy. I don't think it should lead into a bug if the developer understands what returning a pointer mean.
Regarding returning only the index, I considered so and I didn't have any strong feeling against any choice, just found this one more flexible.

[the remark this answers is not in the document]
bmozaffa: But that's my point. You're returning a pointer to a copy! :-)

Create a unit test and give it a try!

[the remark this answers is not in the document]
ruromero: Indeed, the function is receiving an array of copies. Thanks for the correction. Shouldn't we be passing arrays of pointers to avoid creating too many copies?

PR #23, opened 2018-12-04
Added volumes test and improved triggers merge

ruromero:

Added more specific tests for deploymentConfigs
Added failing tests for volumes. I'm afraid that all the arrays will require custom transformers or merge functions.

* Volumes **\***
* InitContainers --> Same as containers
* Containers
  * Command --> simply override, nothing to do
  * Args --> simply override, nothing to do
  * Ports --> Already done
  * EnvFrom
  * Env --> Already done
  * VolumeMounts **\***
  * VolumeDevices
* ImagePullSecrets **\***
* Tolerations
* HostAliases
* ReadinessGates

**\*** I understand we shouldn't try to cover all the possible cases but at least the ones that are more likely to be overridden.

### DIFF
--- a/pkg/controller/kieapp/defaults/merge.go
+++ b/pkg/controller/kieapp/defaults/merge.go
@@ -3,6 +3,7 @@ package defaults
 import (
 	"reflect"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/imdario/mergo"
 	"github.com/kiegroup/kie-cloud-operator/pkg/apis/app/v1"
 	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp/shared"
@@ -233,8 +234,8 @@ func mergeTemplate(baseline *corev1.PodTemplateSpec, overwrite *corev1.PodTempla
 func mergeTriggers(baseline appsv1.DeploymentTriggerPolicies, overwrite appsv1.DeploymentTriggerPolicies) (appsv1.DeploymentTriggerPolicies, error) {
 	var mergedTriggers []appsv1.DeploymentTriggerPolicy
 	for baselineIndex, baselineItem := range baseline {
-		_, found := findDeploymentTriggerPolicy(baselineItem, overwrite)
-		if found == (appsv1.DeploymentTriggerPolicy{}) {
+		idx, found := findDeploymentTriggerPolicy(baselineItem, overwrite)
+		if idx == -1 {
 			logrus.Debugf("Not found, adding %v to slice\n", baselineItem)
 		} else {
 			logrus.Debugf("Will merge %v on top of %v\n", found, baselineItem)
@@ -257,8 +258,8 @@ func mergeTriggers(baseline appsv1.DeploymentTriggerPolicies, overwrite appsv1.D
 		mergedTriggers = append(mergedTriggers, baseline[baselineIndex])
 	}
 	for overwriteIndex, overwriteItem := range overwrite {
-		_, found := findDeploymentTriggerPolicy(overwriteItem, mergedTriggers)
-		if found == (appsv1.DeploymentTriggerPolicy{}) {
+		idx, _ := findDeploymentTriggerPolicy(overwriteItem, mergedTriggers)
+		if idx == -1 {
 			logrus.Debugf("Not found, appending %v to slice\n", overwriteItem)
 			mergedTriggers = append(mergedTriggers, overwrite[overwriteIndex])
 		}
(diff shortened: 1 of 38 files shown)


pkg/controller/kieapp/defaults/merge_test.go
```diff
@@ -279,11 +285,68 @@ func TestMergeDeploymentconfigs_Spec(t *testing.T) {
...
+		*buildDC("dc1"),
+	}
+	overwrite[0].Spec.Template.Spec.Volumes = append(overwrite[0].Spec.Template.Spec.Volumes, corev1.Volume{
+		Name: "dc1-other-volume",
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	})
+	overwrite[0].Spec.Template.Spec.Volumes = append(overwrite[0].Spec.Template.Spec.Volumes, corev1.Volume{
+		Name: "dc1-secret-volume",
```
bmozaffa: This seems like an invalid test. You are adding a duplicate volume name in the same template spec.

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -210,8 +208,8 @@ func mergeTemplate(baseline *corev1.PodTemplateSpec, overwrite *corev1.PodTempla
 func mergeTriggers(baseline appsv1.DeploymentTriggerPolicies, overwrite appsv1.DeploymentTriggerPolicies) (appsv1.DeploymentTriggerPolicies, error) {
 	var mergedTriggers []appsv1.DeploymentTriggerPolicy
 	for baselineIndex, baselineItem := range baseline {
-		_, found := findDeploymentTriggerPolicy(baselineItem, overwrite)
-		if found == (appsv1.DeploymentTriggerPolicy{}) {
+		idx, found := findDeploymentTriggerPolicy(baselineItem, overwrite)
+		if idx == -1 {
```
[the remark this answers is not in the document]
ruromero: Purely Cosmetic

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -251,10 +249,19 @@ func mergeImageChangeParams(baseline appsv1.DeploymentTriggerImageChangeParams,
 	return baseline, nil
 }
 
+// findDeploymentTriggerPolicy Find a deploymentTrigger by Type. In case type == ImageChange
+// the match will be returned if both are not empty
```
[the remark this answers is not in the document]
ruromero: Yes, when the strategy is SourceStrategy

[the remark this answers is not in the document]
bmozaffa: Hmm, ok, and what's the scenario you are trying to cover, and why?

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -171,13 +171,11 @@ func mergeSpec(baseline appsv1.DeploymentConfigSpec, overwrite appsv1.Deployment
 		return appsv1.DeploymentConfigSpec{}, err
 	}
 	overwrite.Template = mergedTemplate
-
```
bmozaffa: I think these empty lines improve readability.

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -171,13 +171,11 @@ func mergeSpec(baseline appsv1.DeploymentConfigSpec, overwrite appsv1.Deployment
...
 	mergedTriggers, err := mergeTriggers(baseline.Triggers, overwrite.Triggers)
 	if err != nil {
 		return appsv1.DeploymentConfigSpec{}, err
 	}
 	overwrite.Triggers = mergedTriggers
-
```
bmozaffa: Ditto

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -251,10 +249,19 @@ func mergeImageChangeParams(baseline appsv1.DeploymentTriggerImageChangeParams,
...
 
+// findDeploymentTriggerPolicy Find a deploymentTrigger by Type. In case type == ImageChange
+// the match will be returned if both are not empty
 func findDeploymentTriggerPolicy(object appsv1.DeploymentTriggerPolicy, slice []appsv1.DeploymentTriggerPolicy) (int, appsv1.DeploymentTriggerPolicy) {
+	emptyImageChangeParams := &appsv1.DeploymentTriggerImageChangeParams{}
 	for index, candidate := range slice {
 		if candidate.Type == object.Type {
-			return index, candidate
+			if object.Type == appsv1.DeploymentTriggerOnImageChange {
+				if !reflect.DeepEqual(object.ImageChangeParams, emptyImageChangeParams) && !reflect.DeepEqual(candidate.ImageChangeParams, emptyImageChangeParams) {
```
bmozaffa: I believe a simple "!=" should suffice here

pkg/controller/kieapp/defaults/merge.go
bmozaffa: Also, we've had a lot of updates, so please rebase. You may be better off manually applying your changes on top of the latest (on this same PR/commit)

pkg/controller/kieapp/defaults/merge.go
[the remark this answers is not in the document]
ruromero: If two triggers of type ImageChange are present only the one that is not empty should be merged.
Currenlty it is not possible to _remove_ triggers or any other object that doesn't allow annotations

pkg/controller/kieapp/defaults/merge_test.go
ruromero: You're right, thanks!

pkg/controller/kieapp/defaults/merge.go
ruromero: No, it is not enough. It is failing to resolve as equal the case where the baseline has an empty ImageChangeParams and the override overrides the the image name.
Actually I preferred to use the go-cmp library instead of the reflect package.

pkg/controller/kieapp/defaults/merge.go
ruromero: Done

pkg/controller/kieapp/defaults/merge.go
bmozaffa: Can you please create a unit test method for this? I want to try locally changing it to "!=" and see why it doesn't work, cause it should.

PR #24, opened 2018-12-04
[KIECLOUD-53] Leverage ConfigMaps for storing/referencing environment yaml files

tchughesiv:
[KIECLOUD-53] 
 - operator creates configmaps for files that packr loads from "config" dir
 - when an environment is deployed, operator will attempt to load from deployed configmaps in openshift... if none available, the KieApp deployment will fail and retry until said ConfigMap is available.
 - this allows user to modify configs to their liking, via configmap, before deploying a given environment.


Makefile
```diff
@@ -34,7 +34,7 @@ vet: sdk-generate
 	$(Q)go vet ./...
 
 test: vet format
-	$(Q)go test ./...
+	$(Q)GOCACHE=off go test ./...
```
bmozaffa: Is this to force-run tests even when there haven't been code changes? I noticed I wasn't getting tests by just changing yaml files.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -127,26 +131,32 @@ func getEnvTemplate(cr *v1.KieApp) v1.EnvTemplate {
...
-		bytes, err := box.Find(filename)
-		if err != nil {
-			logrus.Error(err)
-			return nil, err
+func loadYaml(client client.Client, filename, namespace string, t v1.EnvTemplate) ([]byte, error) {
+	cmName, file := convertToConfigMapName(filename)
+	logrus.Debugf("Loading contents from %s in ConfigMap '%s'", file, cmName)
+	configMap := &corev1.ConfigMap{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: namespace}, configMap)
+	if err != nil {
```
bmozaffa: Is there a way to distinguish between "I'm running a unit test or don't want configmaps" and "your attempt to create/update a configmap was unsuccessful and there is an error".

Too often, our software ignores errors, leading to poor usability. I'm thinking something went wrong and the customer is scratching their head, wondering why their configmap change isn't working as intended, not realizing it's being ignored in favor of the default one.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -159,3 +169,45 @@ func parseTemplate(e v1.EnvTemplate, objBytes []byte) []byte {
 
 	return b.Bytes()
 }
+
+func convertToConfigMapName(filename string) (configMapName, file string) {
+	name := constants.ConfigMapPrefix
+	result := strings.Split(filename, "/")
+	if len(result) > 1 {
+		name = strings.Join([]string{name, result[0]}, "-")
```
bmozaffa: Am I missing something, or should this be results[len(results) - 1]? I assume you want to handle things like env/trial.yaml, no?

Makefile
tchughesiv: correct

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: this is actually naming the configmap based on the directory the file resides in. that configmap will then contain a map of all the files (in yaml form) that that directory is home to.

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: the create configmap logic resides in the kieapp controller code. it will throw an error if there was an issue reconciling the configmap in question.
https://github.com/kiegroup/kie-cloud-operator/blob/3d7d81ecbc36df26e5b4efd42294bbe47db3750b/pkg/controller/kieapp/kieapp_controller.go#L558-L576

pkg/controller/kieapp/defaults/defaults.go
bmozaffa: I'd like to hear the pros and cons of individual configmaps versus grouping them together. I worry that this would be the worst of all worlds. I assume an individual configmap (per yaml file) is easier to create manually, edit, etc... on the other hand, grouping them together will create less clutter and organize everything neatly into one configmap for the operator. Any other pros/cons?

pkg/controller/kieapp/defaults/defaults.go
bmozaffa: Not sure I follow... I don't see where the error is thrown, but also back to line of code that I commented on... what are (all) the possible scenarios where err is not nil?

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: one of the main reasons i went this direction was configmaps didn't support special characters in the object name. so, things like '/' and '.yaml' were failing to create. the other reason is, this mirrors how the oc cli creates configmaps. e.g.
`oc create configmap kieconfig-envs --from-file config/envs/`

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: its thrown in the `reconciler.createObj` function.

tchughesiv: hmm... so would you like something that specifies whether the CM doesn't exist vs an error on retrieval of said CM?

pkg/controller/kieapp/defaults/defaults.go
bmozaffa: Ok, that's fair... another issue is that in your example above, the code will use "config" for the name, whereas you mean it should be "env".  So I basically understand that the method accepts either a single file, or a file within one directory deep, and that's all we need today. This is not high priority, but ideally we would avoid potential for future bugs, where enhancements require two levels deep directories. I don't think you need to support that now, but should make clear it's not supported. Best if that's done through code, the signature only takes a directory name and/or file name, there are two separate functions, or it just causes an error if file path has more than two parts. It can also just be a comment on top of the function.

tchughesiv: I've enhanced the code to support all levels of directory depth

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: this has been improved as you asked... if a configmap is unavailable, the KieApp deployment will fail and the user will see this in the logs -
`ERRO[0022] tommy5/kieconfigs-envs ConfigMap inaccessible, 'production' KieApp not deployed. Retrying...  `

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -145,10 +145,37 @@ type ReconcileKieApp struct {
...
 
-	// Fetch/Create critical ConfigMaps
-	reconciler.initConfigMaps(request.Namespace)
+	// Create critical ConfigMaps if don't exist
+	configMaps := defaults.ConfigMapsFromFile(request.Namespace)
+	for _, configMap := range configMaps {
+		var testDir bool
+		result := strings.Split(configMap.Name, "-")
+		if len(result) > 1 {
+			if result[1] == "testdata" {
```
bmozaffa: It's not clear to me that this testdata directory will be a viable testing strategy for us. However, it looks like you're also looking at the client object to determine it's a unit test, so it should work for tests relying on our main yaml files too, right?

tchughesiv: correct, works for both. this piece of code just keeps the operator from generating configmaps from these test files.

PR #28, opened 2018-12-10
[KIECLOUD-67] Resolve environment merge issues

tchughesiv:
[KIECLOUD-67]
 - added Role & VolumeMount merge logic
 - added view role for rhpam serviceaccount
 - cleaned up env yaml

### DIFF
--- a/config/envs/production.yaml
+++ b/config/envs/production.yaml
@@ -102,6 +102,10 @@ others:
                       value: "KIE Server Router"
                     - name: KIE_SERVER_ROUTER_PROTOCOL
                       value: ""
+                    - name: KIE_SERVER_CONTROLLER_USER
+                      value: controllerUser
+                    - name: KIE_SERVER_CONTROLLER_PWD
+                      value: "{{.ControllerPassword}}"
                     - name: KIE_SERVER_CONTROLLER_TOKEN
                       value: ""
                     - name: KIE_SERVER_CONTROLLER_SERVICE
@@ -168,6 +172,7 @@ servers:
           resources:
             requests:
               storage: "1Gi"
+
     ## PostgreSQL persistent volume claim END
     ## KIE server deployment config BEGIN
     deploymentConfigs:
--- a/config/envs/production.yaml
+++ b/config/envs/production.yaml
@@ -153,324 +153,324 @@ others:
 
 ## KIE Servers BEGIN
 servers:
-## RANGE BEGINS
-  {{ range $index, $Map := .ServerCount }}
-## PostgreSQL persistent volume claim BEGIN
-- persistentVolumeClaims:
-    - metadata:
-        name: "{{.ApplicationName}}-postgresql-claim-{{$index}}"
-        labels:
-          app: "{{.ApplicationName}}"
-          service: "{{.ApplicationName}}-postgresql-{{$index}}"
-      spec:
-        accessModes:
-          - ReadWriteOnce
-        resources:
-          requests:
-            storage: "1Gi"
-  ## PostgreSQL persistent volume claim END
-  ## KIE server deployment config BEGIN
-  deploymentConfigs:
-    - metadata:
-        name: "{{.ApplicationName}}-kieserver-{{$index}}"
-        labels:
-          app: "{{.ApplicationName}}"
-          service: "{{.ApplicationName}}-kieserver-{{$index}}"
-      spec:
-        replicas: 3
-        template:
-          metadata:
-            name: "{{.ApplicationName}}-kieserver-{{$index}}"
-            labels:
-              deploymentConfig: "{{.ApplicationName}}-kieserver-{{$index}}"
-              app: "{{.ApplicationName}}"
-              service: "{{.ApplicationName}}-kieserver-{{$index}}"
-          spec:
-            serviceAccountName: "{{.ApplicationName}}-rhpamsvc"
-            terminationGracePeriodSeconds: 60
-            containers:
-              - name: "{{.ApplicationName}}-kieserver"
-                image: "rhpam{{.Version}}-kieserver-openshift"
-                imagePullPolicy: Always
-                resources:
-                  requests:
-                    memory: 512Mi
-                  limits:
-                    memory: 1Gi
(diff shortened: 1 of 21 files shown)


config/envs/production.yaml
```diff
@@ -168,6 +172,12 @@ servers:
           resources:
             requests:
               storage: "1Gi"
+
+      - metadata:
+          name: "{{.ApplicationName}}-h2-claim-{{$index}}"
```
bmozaffa: Please review this PVC in all env files. Is one needed in 3+ environments? If not, let's remove it from common.yaml. If yes, it should probably have a better name.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -31,25 +31,32 @@ func GetEnvironment(cr *v1.KieApp, client client.Client) (v1.Environment, error)
 	var common v1.Environment
 	yamlBytes, err := loadYaml(client, "common.yaml", cr.Namespace, envTemplate)
 	if err != nil {
+		logrus.Error(err)
```
bmozaffa: Generally speaking, a function returning an error shouldn't need to log it. It's letting the caller know that it failed, and that's enough. The caller would then either bubble up the error, or handle it. The final caller up the stack handling the error would log it. This would avoid multiple redundant error messages.

Another reason for this is that only the caller knows what's an error and what's not. Imagine that a unit test somehow removed or corrupted common.yaml, expecting this function to return an error. It would then declare success at seeing the error and logging something at the error verbosity would not be appropriate. This may not make much sense for this specific instance, but that's a good reason to do the above.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -31,25 +31,32 @@ func GetEnvironment(cr *v1.KieApp, client client.Client) (v1.Environment, error)
...
 		return v1.Environment{}, err
 	}
 	err = yaml.Unmarshal(yamlBytes, &common)
 	if err != nil {
+		logrus.Debugln(string(yamlBytes))
+		logrus.Error(err)
```
bmozaffa: Same here

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -31,25 +31,32 @@ func GetEnvironment(cr *v1.KieApp, client client.Client) (v1.Environment, error)
...
 		return v1.Environment{}, err
 	}
 
 	var env v1.Environment
 	yamlBytes, err = loadYaml(client, fmt.Sprintf("envs/%s.yaml", cr.Spec.Environment), cr.Namespace, envTemplate)
 	if err != nil {
+		logrus.Error(err)
```
bmozaffa: Same here

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -31,25 +31,32 @@ func GetEnvironment(cr *v1.KieApp, client client.Client) (v1.Environment, error)
...
 		return v1.Environment{}, err
 	}
 	err = yaml.Unmarshal(yamlBytes, &env)
 	if err != nil {
+		logrus.Debugln(string(yamlBytes))
+		logrus.Error(err)
```
bmozaffa: Same here

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -31,25 +31,32 @@ func GetEnvironment(cr *v1.KieApp, client client.Client) (v1.Environment, error)
...
 		return v1.Environment{}, err
 	}
 
 	mergedEnv, err := merge(common, env)
 	if err != nil {
+		logrus.Error(err)
```
bmozaffa: Same here

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -494,9 +506,9 @@ func (reconciler *ReconcileKieApp) createCustomObject(obj v1.OpenShiftObject, cr
 	//TODO temp
 	bytes, err := yaml.Marshal(obj)
 	if err != nil {
-		fmt.Printf("Failed to marshall object %v", err)
+		logrus.Errorf("Failed to marshall object %v", err)
 	} else {
-		fmt.Printf("Will create\n\n%v\n\n", string(bytes))
+		logrus.Debugf("Will create\n\n%v\n\n", string(bytes))
```
bmozaffa: Please remove these 5-6 lines, they were only there for troubleshooting and if everything deploys fine now, they've outlived their usefulness.

config/envs/production.yaml
tchughesiv: its only needed for authoring... ill make the changes.

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: agree

config/envs/production.yaml
tchughesiv: done

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: done

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: done

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: done

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: done

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: done

pkg/controller/kieapp/kieapp_controller.go
tchughesiv: done

bmozaffa:
Thanks, all looks good!

PR #33, opened 2018-12-13
[KIECLOUD-63] Upgrade to sdk 0.3.0

tchughesiv:
[KIECLOUD-63]

- sdk 0.3.0
- kubernetes v1.12.3
- controller-runtime v0.1.8
- Match logging changes made w/ this release:
  - https://github.com/operator-framework/operator-sdk/blob/master/doc/user/logging.md
  - `logrus` no longer a dependency
- Create/Update Object improvements


tchughesiv:
@bmozaffa ready for review

cmd/manager/main.go
```diff
@@ -10,75 +13,97 @@ import (
...
-		logrus.Fatalf("failed to get watch namespace: %v", err)
+		log.Error(err, "failed to get watch namespace")
+		os.Exit(1)
 	}
 
 	// Get a config to talk to the apiserver
 	cfg, err := config.GetConfig()
 	if err != nil {
-		logrus.Fatal(err)
+		log.Error(err, "")
```
bmozaffa: How about: log.Error(err, "Error while getting config")

cmd/manager/main.go
```diff
@@ -10,75 +13,97 @@ import (
...
+		os.Exit(1)
+	}
+
+	// Become the leader before proceeding
+	leader.Become(context.TODO(), "kie-cloud-operator-lock")
+
+	r := ready.NewFileReady()
+	err = r.Set()
+	if err != nil {
+		log.Error(err, "")
```
bmozaffa: Same here, "error on NewFileReady()" or something

cmd/manager/main.go
```diff
@@ -10,75 +13,97 @@ import (
...
+		os.Exit(1)
 	}
+	defer r.Unset()
 
 	// Create a new Cmd to provide shared dependencies and start components
 	syncPeriod := time.Duration(2) * time.Hour
 	mgr, err := manager.New(cfg, manager.Options{Namespace: namespace, SyncPeriod: &syncPeriod})
 	if err != nil {
-		logrus.Fatal(err)
+		log.Error(err, "")
```
bmozaffa: ditto

cmd/manager/main.go
```diff
@@ -10,75 +13,97 @@ import (
...
+		os.Exit(1)
 	}
 
-	logrus.Print("Registering Components.")
+	log.Info("Registering Components.")
 
 	// Setup Scheme for all resources
 	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		logrus.Fatal(err)
+		log.Error(err, "")
```
bmozaffa: Here too

cmd/manager/main.go
```diff
@@ -10,75 +13,97 @@ import (
...
+		os.Exit(1)
 	}
 
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
-		logrus.Fatal(err)
+		log.Error(err, "")
```
bmozaffa: Here too

deploy/operator.yaml
```diff
@@ -15,14 +15,21 @@ spec:
...
             - kie-cloud-operator
           imagePullPolicy: Always
+          readinessProbe:
+            exec:
+              command:
+                - stat
+                - /tmp/operator-sdk-ready
+            initialDelaySeconds: 4
+            periodSeconds: 10
+            failureThreshold: 1
```
bmozaffa: Should we have a similar health probe?

pkg/controller/kieapp/defaults/.packr/packr.go
```diff
@@ -17,6 +19,7 @@ func main() {
 
 	err := b.Run()
 	if err != nil {
-		logrus.Fatal(err)
+		log.Error(err, "")
```
bmozaffa: Add message

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -154,7 +155,7 @@ func ConfigMapsFromFile(namespace string) []corev1.ConfigMap {
 	for _, filename := range box.List() {
 		s, err := box.FindString(filename)
 		if err != nil {
-			logrus.Error(err)
+			log.Error(err, "")
```
bmozaffa: log message for 3 occurrences above

pkg/controller/kieapp/defaults/merge.go
```diff
@@ -562,7 +562,7 @@ func mergeRoutes(baseline []routev1.Route, overwrite []routev1.Route) []routev1.
 		slice := make([]routev1.Route, combinedSize(baselineRefs, overwriteRefs))
 		err := mergeObjects(baselineRefs, overwriteRefs, slice)
 		if err != nil {
-			logrus.Errorf("%v", err)
+			log.Error(err, "")
```
bmozaffa: log message in cases above

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -265,13 +276,15 @@ func (reconciler *ReconcileKieApp) GetClient() client.Client {
...
 func (reconciler *ReconcileKieApp) checkImageStreamTag(name, namespace string) bool {
+	log := log.WithValues("kind", "ImageStreamTag", "name", name, "namespace", namespace)
 	result := strings.Split(name, ":")
 	if len(result) == 1 {
 		result = append(result, "latest")
 	}
 	tagName := fmt.Sprintf("%s:%s", result[0], result[1])
 	_, err := reconciler.imageClient.ImageStreamTags(namespace).Get(tagName, metav1.GetOptions{})
 	if err != nil {
+		log.V(1).Info("Object does not exist")
```
bmozaffa: Is 1 for info? I would think error would be 0, warning would be 1, and info would be 2 or higher? This new logger, btw, leaves much to be desired. I'm thinking maybe we should provide our own layer of indirection.

bmozaffa: So are error and warning considered implicitly known levels and then it starts counting up from 0 for Info? Do you have a link for that?

As far as adding a layer, I'm not suggesting we stop using this logger as you have. I'm suggesting having Error(), Warn(), Info(), and Debug() functions in one Go class and in there, call log.Error or log.V(1) and so on, as appropriate. The benefit would be that if we decide to make a chance, it's just in one place.

tchughesiv: We use the following -
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.1.8/pkg/runtime/log/log.go#L43-L68

`logr` is what is leveraged from the controller-runtime log pkg -
https://github.com/go-logr/logr#differences-from-daves-ideas

tchughesiv: only debug, error, and info are currently recognized levels... stops at 1.

bmozaffa: Got it RE levels... my point stands about the extra layer. Not sure if you meant to respond to that as well.

tchughesiv:
@bmozaffa added messages to all Error logs.

pkg/controller/kieapp/kieapp_controller.go
tchughesiv: did we want to do something with this PR or separately?

bmozaffa: Let's go this PR, because it's the exact same lines being changed here.

cmd/manager/main.go
```diff
@@ -19,26 +19,10 @@ import (
...
-	// implementing the logr.Logger interface. This logger will
-	// be propagated through the whole operator, generating
-	// uniform and structured logs.
-	if debugBool {
-		logf.SetLogger(logf.ZapLogger(true))
-	} else {
-		logf.SetLogger(logf.ZapLogger(false))
-	}
-}
+var log = kieapp.SugaredLogger("cmd")
```
bmozaffa: My concern is that this is not our layer of indirection. It is "zap.SugaredLogger".

Here is a test for whether the layer of indirection solves the problem. You changed around 16 files in order to switch to the new logger. How many files would you need to change if we switched back to the old logger now? Ideally, the answer would be, only our own logger file.

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -198,7 +197,7 @@ func (reconciler *ReconcileKieApp) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, err
 	}
 
-	log := log.WithValues("kind", instance.Kind, "name", instance.Name, "namespace", instance.Namespace)
+	log := log.With("kind", instance.Kind, "name", instance.Name, "namespace", instance.Namespace)
```
bmozaffa: This syntax (or the previous one, WithValues) seems to be one distinguishing factor of the logger. That means adopting it in our layer would bind us to this logger. Question is, how useful is it? I see the log var only being used once below, meaning the same info could have been provided down there. What are your thoughts?

cmd/manager/main.go
tchughesiv: I had to change them to get away from the controller-runtime logger reference in every pkg

tchughesiv: going forward, it would only be changes in our logger class

tchughesiv: but ya, i can change the function name to something more generic?

pkg/controller/kieapp/kieapp_controller.go
tchughesiv: i find it very useful... but give it a spin and offer your thoughts

deploy/operator.yaml
bmozaffa: Any thoughts on liveness probe? Does OpenShift warn you when one is missing? Can the same readiness probe be used for liveness? Should it?

cmd/manager/main.go
bmozaffa: It's not about the function name. The interface you return is "zap.SugaredLogger", which may be ok. It just means that if we switch to another logger implementation, it will still be a lot of work. we would have to implement some 20+ functions and delegate to the new logger. I'm ok with crossing that bridge when we get to it, though.

pkg/controller/kieapp/kieapp_controller.go
bmozaffa: Sounds good... I can imagine it being useful, but I assume only if you create a logger with the properties and then log with it several times.

Let me know about the probe before I merge this PR.

tchughesiv: true, ill look at whether we are better off passing in the args as needed

bmozaffa: I meant about the liveness probe! :-)

deploy/operator.yaml
tchughesiv: can we think a little more about the liveness probe? i'm not sure this script is the best way to go about it -
https://github.com/operator-framework/operator-sdk/blob/dba3d18eec0bada10d13cbe7ba474df5f3e99321/pkg/ready/ready.go#L45

The readiness probe is a good start.

tchughesiv: it appears to just create the file and the readiness probe then ensures it exists ... not a great way to approach liveness

cmd/manager/main.go
tchughesiv: I've at least changed the function name to something more generic to make it easier going forward and ya, I'm sure can be improved. Let's get these sdk upgrades merged so we can iterate around logging and other items we have.

pkg/controller/kieapp/logs/logs.go
```diff
@@ -0,0 +1,102 @@
...
+	if debugBool {
+		logger = CreateLogger(true)
+	} else {
+		logger = CreateLogger(false)
+	}
+	logger.Logger = logf.Log.WithName(name)
+	return logger.SugaredLogger.Named(name)
+}
+
+func CreateLogger(development bool) (logger Logger) {
```
bmozaffa: Should these 4 functions be exported? I assumed only the first one would be, and the rest would be internal implementation and subject to change

tchughesiv: agree

tchughesiv: i've changed 2 of the function names, however the GetEnv functions still need to be exported as they are called from other packages.

bmozaffa: Should we move those 2 up the shared.utils or somewhere?

tchughesiv: that's actually where they existed previously, and it was my preference. the problem is, i was hitting import conflicts when trying to use those functions from logs. we need to use them to check for the DEBUG env var from GetLogger... and since the shared/utils packages call the logs package for logging, those functions can't exist in shared/utils. 

we could offer the functions in both locations and only export from shared location... but felt didn't want to duplicate code.

PR #38, opened 2019-01-06
[KIECLOUD-86] Use local namespace for objects after creating image streams

bmozaffa:
When the image stream is created locally, the local namespace should be  returned to be set on the object. Also refactored function to be more clear.


bmozaffa:
@tchughesiv @ruromero FYI

PR #41, opened 2019-01-07
[KIECLOUD-89] Refactor operator code for readability and simplicity

bmozaffa:
Refactor of code to include mostly boilerplate in top package, and clean up kieapp_controller a bit.

Also introduced further abstraction for platform services to allow unit testing of all parts of code.


bmozaffa:
@tchughesiv @ruromero Please review and provide feedback. I'm hoping there is more we can do, and also that this bit is helpful.

pkg/controller/initializer.go
```diff
@@ -0,0 +1,76 @@
...
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	addManager := func(mgr manager.Manager) error {
+		k8sService := GetInstance(mgr)
+		reconciler := kieapp.KieAppReconciler{Service: &k8sService}
+		return add(mgr, &reconciler)
+	}
+	AddToManagerFuncs = []func(manager.Manager) error{addManager}
+}
+
+func add(mgr manager.Manager, reconciler reconcile.Reconciler) error {
```
tchughesiv: does moving this up one level make it more difficult to add future controllers? I feel like the way its currently organized we have more flexibility around what each controller watches/loads. i know we currently only have 1 (kieapp) but maybe in the future we offer an upgrade controller or something?

bmozaffa: hmm, is it likely that we'll add another controller? We could move the initializer.go file one level down, but the init() will have to stay in this package to happen in time, and we'll end up with a Go file with one/two lines in it again.

I was trying to put myself in the shoes of someone reading the code for the first/second time. It's really hard to understand what happens and in what order, even for me sometimes. Maybe it's ok to have an initializer both up here and down there, even if the top one only has a few lines. WDYT?

tchughesiv: i think that might better, so we can confine the controller-specific watches/logic to each controller package the operator might offer in the future. thoughts?

tchughesiv:
LGTM

PR #43, opened 2019-01-08
[KIECLOUD-89] Log messages should be in lowercase. Refactor pwd gener…

ruromero:
* Log messages in Go should start with lower case
* Password generation logic extracted to a new function to reduce duplicity

Fix https://issues.jboss.org/browse/KIECLOUD-89

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -88,11 +88,7 @@ func getEnvTemplate(cr *v1.KieApp) v1.EnvTemplate {
 		}
 	}
 
-	setPassword(&config.KeyStorePassword, isTrialEnv)
-	setPassword(&config.AdminPassword, isTrialEnv)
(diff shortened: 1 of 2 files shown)


tchughesiv:
@ruromero do we have a test that checks if the operator keeps a password that is already set. if not, can you add one? otherwise LGTM.

ruromero:
@tchughesiv these 2 tests should be covering that part https://github.com/kiegroup/kie-cloud-operator/blob/1204e124045ef4a907c1b0b5c590973e97718169/pkg/controller/kieapp/defaults/defaults_test.go#L330-L379

bmozaffa:
> * Log messages in Go should start with lower case

Source / reference for this rule/convention?

ruromero:
It seems it only applies to error messages
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

ruromero:
@bmozaffa done

PR #45, opened 2019-01-09
[KIECLOUD-91] Implement auth integration

ruromero:

Fix https://issues.jboss.org/browse/KIECLOUD-91

Added sso/ldap/roleMapper model
Added conditional envVars to common template
Added unit tests for common cases

**Notes**: 
* ~~I don't know why I needed to declare $kieDeployments and $auth variables in common.yaml (https://github.com/kiegroup/kie-cloud-operator/commit/65ff909dcccb6ed37a40e27b56c43689816a57f3#diff-fce920fb56cc0a73c2b47a3e4ed9a3dcR389)~~
* ~~Why existing template statements are commented? e.g. range~~
* ~~Should I propagate the routes (or through some kind of context variable) down to the getEnvironment method? Or instead add it in the kieapp_controller before the objects consolidation? (https://github.com/kiegroup/kie-cloud-operator/commit/65ff909dcccb6ed37a40e27b56c43689816a57f3#diff-53394291f02ee59483694ad21df717b5R190)~~


bmozaffa:
> * I don't know why I needed to declare $kieDeployments and $auth variables in common.yaml ([65ff909#diff-fce920fb56cc0a73c2b47a3e4ed9a3dcR389](https://github.com/kiegroup/kie-cloud-operator/commit/65ff909dcccb6ed37a40e27b56c43689816a57f3#diff-fce920fb56cc0a73c2b47a3e4ed9a3dcR389))

Not sure what you're saying/asking. It looks like you need kieDeployments to decided if to insert the statements in line 524?

> * Why existing template statements are commented? e.g. range

I now remember and maybe better understand your question. You are not asking why the range statement is there, but why it's commented out. That's because it's not valid yaml and messes up something... can't remember what!

ruromero:
Thanks @bmozaffa I commented out the template code to make it a valid yaml file
Regarding the need of define `$auth` variable, maybe @tchughesiv might know.
Finally, as discussed online I'll propose an implementation for the route value propagated to the HOSTNAME_HTTPS variable

ruromero:
I'll reply to myself https://golang.org/pkg/text/template/#hdr-Variables

> When execution begins, $ is set to the data argument passed to Execute, that is, to the starting value of dot.

bmozaffa:
> I'll reply to myself https://golang.org/pkg/text/template/#hdr-Variables
> 
> > When execution begins, $ is set to the data argument passed to Execute, that is, to the starting value of dot.

I'm not sure I understand. Why can't you use ".Auth"? I assume with a loop, the $auth represents each iteration value, but you have a constant ".Auth", don't you?

ruromero:
> > I'll reply to myself https://golang.org/pkg/text/template/#hdr-Variables
> > > When execution begins, $ is set to the data argument passed to Execute, that is, to the starting value of dot.
> 
> I'm not sure I understand. Why can't you use ".Auth"? I assume with a loop, the $auth represents each iteration value, but you have a constant ".Auth", don't you?

It seems that the current scope outside of the `range` is saved to `$` and then passed to the scope of the `range`. What I don't get to understand is why `{{.ApplicationName}}` works but `{{.Auth.SSO.URL}}` doesn't ONLY inside the `range::if`

ruromero:
@bmozaffa @tchughesiv it should be ready now

deploy/crs/kieapp_trial_sso.yaml
```diff
@@ -0,0 +1,21 @@
...
+      adminUser: admin
+      adminPassword: example
+      clients:
+        console:
+          name: rhpamcentr-client
+          secret: somePwd
+        servers:
+          - name: kieserver-client-a
+            secret: otherPwd
+            hostnameHTTPS: kieserver-a.example.com
```
tchughesiv: what's this used for? is it required to specify the route here?

ruromero: This will be used by the EAP startup scripts to create a RHSSO Client in the realm with this redirectURL. If not set, the route hostname will be used instead (only when the client is set)

tchughesiv: ok, so its optional? if so, should we include it in our example repo? Seems like this would break in most clusters unless they're resolving example.com locally?

ruromero: This was just an example configuration. I better remove the file completely

tchughesiv: idk, might be good to have, but maybe just include the required fields? the user would see the optional fields post deployment... right?

ruromero: Moved to the README.md file

pkg/controller/kieapp/defaults/auth.go
```diff
@@ -0,0 +1,93 @@
...
+	} else if spec.Auth.Type == v1.LDAPAuthType {
+		err = configureLDAP(spec.Auth.LDAP, envTemplate)
+	} else {
+		return fmt.Errorf("invalid authentication type %v", spec.Auth.Type)
+	}
+	if spec.Auth.RoleMapper != nil {
+		err = configureRoleMapper(spec.Auth.RoleMapper, envTemplate)
+	}
+	return err
+}
```
bmozaffa: How about this, specially for this case where you have a lot of ifs, and declare an error upfront:

```
func configureAuth(spec v1.KieAppSpec, envTemplate *v1.EnvTemplate) (err error) {
	switch {
	case len(spec.Auth.Type) == 0:
		err = nil
	case spec.Auth.Type == v1.SSOAuthType:
		err = configureSSO(spec.Auth.SSO, spec.KieDeployments, envTemplate)
	case spec.Auth.Type == v1.LDAPAuthType:
		err = configureLDAP(spec.Auth.LDAP, envTemplate)
	default:
		err = fmt.Errorf("invalid authentication type %v", spec.Auth.Type)
	}

	if err == nil && spec.Auth.RoleMapper != nil {
		err = configureRoleMapper(spec.Auth.RoleMapper, envTemplate)
	}
	return
}
```

pkg/controller/kieapp/defaults/auth.go
```diff
@@ -0,0 +1,93 @@
...
+		return errors.New("missing LDAP configuration")
+	}
+	if len(config.URL) == 0 {
+		return errors.New("the url must not be empty")
+	}
+	envTemplate.Auth.LDAP = *config.DeepCopy()
+	return nil
+}
+
+func configureRoleMapper(config *v1.RoleMapperAuthConfig, envTemplate *v1.EnvTemplate) error {
```
bmozaffa: Are you just trying to standardize on returning an error? Not sure if that's worth it. Result is that you have an extra check on the calling function, or in your case above, you overwrite a potential error with nil.

pkg/apis/app/v1/kieapp_types.go
```diff
@@ -22,11 +23,12 @@ type KieAppSpec struct {
...
-	KieDeployments int            `json:"kieDeployments"` // Number of KieServer DeploymentConfigs (defaults to 1)
-	RhpamRegistry  KieAppRegistry `json:"rhpamRegistry,omitempty"`
-	Objects        KieAppObjects  `json:"objects,omitempty"`
-	CommonConfig   CommonConfig   `json:"commonConfig,omitempty"`
+	Environment    string           `json:"environment,omitempty"`
+	KieDeployments int              `json:"kieDeployments"` // Number of KieServer DeploymentConfigs (defaults to 1)
+	RhpamRegistry  KieAppRegistry   `json:"rhpamRegistry,omitempty"`
+	Objects        KieAppObjects    `json:"objects,omitempty"`
+	CommonConfig   CommonConfig     `json:"commonConfig,omitempty"`
+	Auth           KieAppAuthObject `json:"auth,omitempty"`
```
bmozaffa: Please also add CRD validation. See my [commit here](https://github.com/kiegroup/kie-cloud-operator/pull/44/commits/68b111b9e4c5e6f2ee062baf050457a96b557f5f), including comments about what's supported today.

README.md
```diff
@@ -1,44 +1,55 @@
...
+
 ```shell
 oc create -f deploy/crs/kieapp_trial.yaml
 ```
 
-# Clean up a KieApp deployment:
+## Clean up a KieApp deployment
+
 ```shell
 oc delete KieApp trial
```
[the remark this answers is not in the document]
ruromero: I was just formatting the existing README according to the markdownlint suggestions but I will do the changes you suggested.

deploy/crs/kieapp_trial_sso.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: app.kiegroup.org/v1
+kind: KieApp
+metadata:
+  name: trial
+spec:
+  environment: trial
+  auth:
+    type: sso
+    sso:
```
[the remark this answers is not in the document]
ruromero: I followed this pattern because I saw something similar in other core objects but it is not necessary. I am removing it

pkg/apis/app/v1/kieapp_types.go
```diff
@@ -125,6 +127,82 @@ type WebhookSecret struct {
...
 
+// KieAppAuthType authentication type
+type KieAppAuthType string
+
+const (
+	// SSOAuthType SSO Authentication type
+	SSOAuthType KieAppAuthType = "sso"
+	// LDAPAuthType LDAP based authentication
+	LDAPAuthType KieAppAuthType = "ldap"
+)
```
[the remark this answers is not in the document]
ruromero: Removed

pkg/controller/kieapp/defaults/auth.go
ruromero: After removing the type I don't think see how to avoid the use of the ifs. I have however declared the error as suggested.

pkg/controller/kieapp/defaults/auth.go
ruromero: Fixing

pkg/apis/app/v1/kieapp_types.go
ruromero: Done

pkg/controller/kieapp/defaults/auth.go
```diff
@@ -0,0 +1,87 @@
...
+					Value: hostname,
+				})
+			} else if len(container.Env[pos].Value) == 0 {
+				container.Env[pos].Value = hostname
+			}
+		}
+	}
+}
+
+func configureAuth(spec v1.KieAppSpec, envTemplate *v1.EnvTemplate) (err error) {
```
bmozaffa: When you name your return variable, you no longer have to return it in the function. You just need to set it to a value and do a plain return. Ideally and in the case of if/elseif, that would be:
```
if() {
err = ...
} else if() {
err = ...
} else {
err = ...
}
return
```

If that no longer makes sense because of how conditions changed, maybe best to just remove the named return variable because it's not being used in this function today.

deploy/crds/kieapp_crd.yaml
```diff
@@ -74,11 +74,11 @@ spec:
...
-                      anyOf:
-                      - required:
-                        - limits
-                      - required:
-                        - requests
+                      # anyOf:
+                      # - required:
+                      #   - limits
+                      # - required:
+                      #   - requests
```
bmozaffa: Did you have a problem with these? I had tested these and despite the fact that the top ones caused problems, these seemed to work just fine with oc 3.11

deploy/crds/kieapp_crd.yaml
```diff
@@ -206,3 +206,167 @@ spec:
...
+                    principalAttribute:
+                      type: string
+                      description: RH-SSO Principal Attribute to use as username
+                    clients:
+                      type: object
+                      description: Client definitions used for creating the RH-SSO clients in the specified Realm
+                      properties:
+                        console:
+                          type: object
+                          description: Console client definition
```
bmozaffa: If you have console in there, I imagine some of the fields should be required?

deploy/crds/kieapp_crd.yaml
```diff
@@ -206,3 +206,167 @@ spec:
...
+                              format: password
+                              description: Client secret
+                            hostnameHTTP:
+                              type: string
+                              description: Hostname to set as redirect URL
+                            hostnameHTTPS:
+                              type: string
+                              description: Secure hostname to set as redirect URL
+                        servers:
+                          type: array
```
bmozaffa: minItems: 1?

deploy/crds/kieapp_crd.yaml
ruromero: I'm not sure because having the console client defined, although empty, will have the controller set the hostnameHTTPS. The client name is generated by the scripts and the secret can be generated by the RH-SSO.

ruromero: I take that back. I think it's better to set the client name mandatory.

ruromero:
@bmozaffa I am having problems with the schema definition. See my commented code. When I create the CR, the controller complains about missing required bits. Are you experiencing this problems when creating, let's say, the trial env with or without a simple auth config?

```
  auth:
    sso:
      url: https://mysso.xample.com
      realm: therealm
      adminUser: test
```

README.md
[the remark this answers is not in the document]
bmozaffa: Yes, thanks, I realized that, but was just asking to fix that while at it :-)

deploy/crds/kieapp_crd.yaml
bmozaffa: With the lab outage, it's difficult to properly validate these. However, I've tested locally and at least with kubectl 1.13 and oc 3.11, the local side validation passes for the trial env with your items uncommented. I think we have the oc 3.10 client in the lab, which I should also try. Please also let me know what you are using, because the version of your client-side oc library is probably what matters most.

pkg/controller/kieapp/defaults/auth.go
bmozaffa: FWIW, in Go, you can use switch/case in the format we usually have to use if/else in Java. It's not necessarily or always better, but can something be cleaner. That's the case for my example above, as I'm not "switching on anything". You can in fact still do so, though it doesn't change much.

bmozaffa:
> @bmozaffa I am having problems with the schema definition. See my commented code. When I create the CR, the controller complains about missing required bits. Are you experiencing this problems when creating, let's say, the trial env with or without a simple auth config?

Ok, so the only remaining issue, IMO, is the commented out schema bits. With the lab down, it will have to wait till Monday. All these work just fine with k8s 1.13, whether with an oc 3.11 or kubectl 1.13 client, but let's test against OCP while making note of the client library/version used.

PR #46, opened 2019-01-13
[KIECLOUD-91] Follow-up PR for CRD validation

bmozaffa:
Commenting out most conditional required elements to allow things to work with OCP 3.11, to be revisited later

It appears that in 3.11, the validation looks one level deeper than what is provided in the CR, resulting in a lot of false positives


bmozaffa:
@ruromero Commenting out most required elements for now. See commit message for more detail.

PR #47, opened 2019-01-14
[KIECLOUD-92] Added unit tests to validate CRs against CRD

bmozaffa:
Provide unit tests to validate distributed CRs against provided CRD and further test CRD validation


bmozaffa:
@tchughesiv @ruromero Please review! We can write a lot more unit tests, but one important function, IMO, is [the one](https://github.com/kiegroup/kie-cloud-operator/pull/47/files#diff-9288d9044393689673f0732ad5124f61R25) that validates the CRs we distribute against our CRD.

Also, I'm not seeing any noticeable image size increase from the previous build. That might be a consideration, so let me know if you find otherwise.

ruromero:
LGTM

tchughesiv:
slick & LGTM. size difference of builds is negligible to me as well.

PR #49, opened 2019-01-16
Test PR to validate CI

ruromero:


build/Dockerfile
```diff
@@ -1,7 +1,19 @@
...
+#
 
-RUN apk upgrade --update --no-cache
+# Build the manager binary
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 as builder
 
-USER nobody
+# Copy in the go src
+WORKDIR /go/src/github.com/kiegroup/kie-cloud-operator
+COPY . .
```
tchughesiv: we should probably add a `RUN NO_DOCKER=1 make build` after this line

build/Dockerfile
```diff
@@ -1,7 +1,19 @@
...
 
-ADD build/_output/bin/kie-cloud-operator /usr/local/bin/kie-cloud-operator
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o kie-cloud-operator github.com/kiegroup/kie-cloud-operator/cmd/manager
+
+# Copy the controller-manager into a thin image
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
```
tchughesiv: this `FROM` should be from a minimal supported image? maybe rhel-atomic? worst case rhel7?

build/Dockerfile
```diff
@@ -1,7 +1,19 @@
...
+WORKDIR /root/
```
tchughesiv: this should be `/usr/local/bin/` instead of root

build/Dockerfile
```diff
@@ -1,7 +1,19 @@
...
+COPY --from=builder /go/src/github.com/kiegroup/kie-cloud-operator/kie-cloud-operator .
+ENTRYPOINT ["./kie-cloud-operator"]
```
tchughesiv: delete this line... the command logic is handled within the operator deployment
`USER nobody` should remain so we aren't running as uid 0

build/Dockerfile
droslean: You are already building on top of this image. see [here](https://github.com/openshift/release/blob/master/ci-operator/config/kiegroup/kie-cloud-operator/kiegroup-kie-cloud-operator-master.yaml#L10)

You don't need this.

build/Dockerfile
```diff
@@ -1,7 +1,19 @@
...
+# Dockerfile for building local images.
+#
 
-RUN apk upgrade --update --no-cache
+# Build the manager binary
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 as builder
 
-USER nobody
+# Copy in the go src
+WORKDIR /go/src/github.com/kiegroup/kie-cloud-operator
```
droslean: You also don't need that. Your code will be cloned from `ci-operator`

build/Dockerfile
```diff
@@ -1,7 +1,19 @@
-FROM alpine:3.8
+#
+# Dockerfile for building local images.
+#
 
-RUN apk upgrade --update --no-cache
+# Build the manager binary
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 as builder
```
droslean: You can also remove this. You can make a combination of the image layers that you want in your ci-operator configuration.

tchughesiv:
@ruromero whatever we ultimately decide the Dockerfile should look like, it should probably work both within the CI system and locally on one's desktop outside of the corporate network.

build/Dockerfile
tchughesiv: hmm, ok... is this a requirement @droslean ? that image is rather large... we'd like to keep our resulting deployable operator image under 75mb.

ruromero: I'll change it by fedora-minimal:30

build/Dockerfile
```diff
@@ -1,7 +1,5 @@
-FROM alpine:3.8
-
-RUN apk upgrade --update --no-cache
+FROM fedora-minimal:30
```
tchughesiv: this should be `FROM registry.fedoraproject.org/fedora-minimal:30` to ensure the image is pulled from the proper location.

ruromero: correct, thanks

hack/go-vet.sh
```diff
@@ -4,4 +4,4 @@ if [[ -z ${CI} ]]; then
     source hack/go-dep.sh
     operator-sdk generate k8s
 fi
-go vet ./...
\ No newline at end of file
+go vet ../...
```
tchughesiv: this doesn't look right... should it be `go vet ./...`?

ruromero: isn't it relative to the `hack` folder?

tchughesiv: ok, i haven't tested... checking.

tchughesiv: based on my testing, `go vet ./...` works as expected. that logic would also w/ the fact that you are issuing a relative `source hack/go-dep.sh` just above this.

droslean:
well, you've got the picture.

tchughesiv:
@droslean i used to be member. so, how then does does the PR get the lgtm label?

PR #52, opened 2019-01-21
[KIECLOUD-95] adding CSV and package for OLM support

tchughesiv:
Compatibility w/ Operator Lifecycle Manager
 - incremented operator version
 - added consoleHost to CR Status
 - added CSV for current version
 - added package for OLM compat
 - hack/catalog.sh will configure OCP OLM catalog for csv/operator use.

Developed/tested against OCP 3.11 w/ OLM tech preview.


deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,201 @@
...
+        statusDescriptors:
+          - description: The status of each of the Deployments for the KieApp environment.
+            displayName: Deployment Status
+            path: deployments
+            x-descriptors:
+              - "urn:alm:descriptor:com.tectonic.ui:podStatuses"
+  install:
+    strategy: deployment
+    spec:
+      permissions:
```
ruromero: Wouldn't it be better if we had a `csv-gen.sh` script that would copy the contents of the permissions and deployments from the standalone yaml files? We would avoid duplicity

deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,201 @@
...
+                - servicemonitors
+              verbs:
+                - get
+                - create
+      deployments:
+        - name: kie-cloud-operator
+          spec:
+            replicas: 1
+            strategy:
+              type: Recreate
```
ruromero: Do you mind moving the strategy.type from spec.template.spec to spec in the operator.yaml file as it is here?

deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,201 @@
...
+            displayName: Deployment Status
+            path: deployments
+            x-descriptors:
+              - "urn:alm:descriptor:com.tectonic.ui:podStatuses"
+  install:
+    strategy: deployment
+    spec:
+      permissions:
+        - serviceAccountName: kie-cloud-operator
+          rules:
```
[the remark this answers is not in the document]
tchughesiv: should yes... but OLM doesn't currently support all of the verbs we require so I had to use more wide-ranging perms to get it working. i don't want to make that a permanent change though so keeping the roles.yaml the same.

deploy/crds/kieapp_crd.yaml
```diff
@@ -291,8 +291,8 @@ spec:
                           items:
                             type: object
                             description: Single Server client definition
-#                            required:
-#                            - name
+                            #                            required:
+                            #                            - name
```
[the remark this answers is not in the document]
tchughesiv: this appears to be a format fix that wasn't applied in a prior PR. I can revert this though.

deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
tchughesiv: ya, good catch

deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
tchughesiv: i think so, all but the roles/rules as there are differences currently

deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,201 @@
...
+  maintainers:
+    - name: Red Hat, Inc.
+      email: bsig-cloud@redhat.com
+  provider:
+    name: Red Hat, Inc.
+  links:
+    - name: Product Page
+      url: https://access.redhat.com/products/red-hat-process-automation-manager
+    - name: Documentation
+      url: https://access.redhat.com/documentation/en-us/red_hat_process_automation_manager/7.1/html-single/installing_and_configuring_red_hat_process_automation_manager_on_red_hat_jboss_eap_7.1/
```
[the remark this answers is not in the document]
tchughesiv: @bmozaffa thoughts?

deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,201 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: kiecloud-operator.v0.0.1
+  # namespace: placeholder
+  annotations:
+    categories: A list of comma separated categories that your operator falls under.
+    certified: The operator's certification. This field should be set to false until a process to get this value is defined at a later date.
+    description: KieCloud can deploy RHPAM environments in the form of KieApp objects.
+    containerImage: quay.io/kiegroup/kie-cloud-operator:latest
```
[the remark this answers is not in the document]
tchughesiv: i'll moved to versioning in this PR

tchughesiv:
> It would be good to add some instructions of how to add it using the `catalog.sh` script to an existing cluster in the `readme.md`

added

deploy/catalog_resources/kiecloud-operator.v0.0.1.clusterserviceversion.yaml
tchughesiv: @bmozaffa @ruromero is this something we want with this PR or a future one?

bmozaffa: I'm eager to get this merged, so I vote for later.

bmozaffa:
> My default example was in the wrong place in the CSV, I've fixed that and it now properly populates `trial`.

Looks good!

> 
> > * I've seen operators offer two/three yaml options on the right hand side, where you can click to populate them into the editor. Any idea what it would take to have that?
> 
> Based on my testing, this appears to be if your operator supports multiple controllers/kinds ... for example, in the future we may offer a KieUpgrade, or a KieDM, vs today only offering KieApp.

Not sure about this. Worth continuing to look into. Maybe that's what it was, but it looked like different CR options to me, and I no longer have the video in question.

PR #60, opened 2019-02-01
[KIECLOUD-102] Support DM deployments

ruromero:

Fix https://issues.jboss.org/browse/KIECLOUD-102

* Renamed existing environment configs from `<environment>.yaml` to `rhpam-<environment>.yaml`
* Renamed existing CRs and created new ones for RHDM
* Created new sets of environments for RHDM
  * rhdm-authoring
  * rhdm-authoring-ha
  * rhdm-production-immutable
  * rhdm-trial
  * rhdm-optaweb-trial
* Moved kieserver persistence config to rhpam templates
* Remove `rhpam` `businesscentral` `RHMAPCENTR` references in config yaml files
* Fixed merge case where there's no need to override the Servers


ruromero:
/hold
I need to test OLM integration but review is welcome

pkg/apis/app/v1/kieapp_types.go
```diff
@@ -23,14 +23,49 @@ type KieAppSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// KIE environment type to deploy (prod, authoring, trial, etc)
-	Environment    string           `json:"environment,omitempty"`
+	Environment    EnvironmentType  `json:"environment,omitempty"`
```
bmozaffa: Mostly a question about the pros/cons of this. I think the advantage, type safety, is obvious and I like it. However, did we previously have an option for users to add their own ad-hoc named environments, and do we lose that?

I can see in the rest of the PR that naming environments brings a lot of benefits, and think it's absolutely acceptable to make this trade-off, but would just want us to make the decision consciously.

bmozaffa:
Great work, @ruromero !

bmozaffa:
FWIW, I tried to test this and there seem to be some issues. I have your build image pushed to quay/kiegroup with a tag of rhdm, and used it through OLM to create an operator in https://console.app.two.ocp.cloud.lab.eng.bos.redhat.com/k8s/ns/rhdm/clusterserviceversions. Creating resources fails, saying the configmap has not been created. Odd, because I didn't notice any relevant changes in the PR. Don't have time to troubleshoot, so definitely give it a look!

pkg/apis/app/v1/kieapp_types.go
ruromero: As it is still a string, it is possible to add more, see the unit tests

ruromero:
@bmozaffa I have tested it myself and it works, the error is normal after deleting the configmaps, this is not related to this PR.

/hold cancel

ruromero:
fixed @tchughesiv please review latest commit

PR #63, opened 2019-02-14
[KIECLOUD-116] Allow immutable environments to use other images

ruromero:

Fix https://issues.jboss.org/browse/KIECLOUD-116


deploy/crds/kieapp.crd.yaml
```diff
@@ -97,45 +93,261 @@ spec:
...
-#                    - resources
                   properties:
-                    env:
-                      type: array
-                      items:
+                    deployments:
+                      type: integer
+                      format: int16
+                      description: Number of Server sets that will be deployed
+                    from:
```
bmozaffa: Is this optional, and does it default to "from RH kie server images"?

deploy/crds/kieapp.crd.yaml
```diff
@@ -97,45 +93,261 @@ spec:
                 server:
```
bmozaffa: We have at least a naming problem here. Maybe can be solved by bringing "servers" under this "server"... maybe calling it something else, or maybe leave as servers

deploy/crds/kieapp.crd.yaml
```diff
@@ -97,45 +93,261 @@ spec:
...
+                    from:
+                      type: object
+                      description: Image definition to use for all the servers
+                      required:
+                        - kind
+                        - name
+                      properties:
+                        kind:
+                          type: string
+                          description: Object kind. e.g. ImageStreamTag
```
bmozaffa: Any other examples? Should this be an enum?

deploy/crds/kieapp.crd.yaml
```diff
@@ -35,10 +35,6 @@ spec:
                 - rhpam-production-immutable
                 - rhpam-production
                 - rhpam-trial
-            kieDeployments:
-              type: integer
-              description: The number of distinct KIE server deployments
-              minimum: 1
```
bmozaffa: I'm thinking this should be optional if they provide a servers array

deploy/crds/kieapp.crd.yaml
```diff
@@ -97,45 +93,261 @@ spec:
...
+                        kind:
+                          type: string
+                          description: Object kind. e.g. ImageStreamTag
+                        name:
+                          type: string
+                          description: Object name
+                        namespace:
+                          type: string
+                          description: Namespace where the object is located
+                    build:
```
bmozaffa: This is to create multiple KIE server sets that are all built from the same git source? I don't think we need this. I suggest removing this, and if they want to do a single s2i build, they can provide a servers array of size one with that in there.

bmozaffa:
Please also rebase!

deploy/crds/kieapp.crd.yaml
ruromero: It is optional. You should either pick the `server` for common config or `servers` for individual one. As you mentioned, a folder with examples is required at this point

deploy/crds/kieapp.crd.yaml
ruromero: Sure, I will limit it to just DockerImage and ImageStreamTag

deploy/crds/kieapp.crd.yaml
ruromero: I decided to let the consumer decide whether to use a common `server` configuration or a specific one for each server.
```
server:
  deployments: 2
```
or
```
servers:
 - name: server1
 - name: server2
```

deploy/crds/kieapp.crd.yaml
ruromero: Yes it is optional and defaults to RH Kies Server images. See 
https://github.com/kiegroup/kie-cloud-operator/pull/63/commits/a0e904a0a46665dcde97eb9aa82eaed262f0cdfb#diff-53394291f02ee59483694ad21df717b5R167-R177

ruromero:
> Looks good over all, but now definitely need an examples folder under deploy with a few sample CRs to better understand our options. I've also asked individual questions that should clarify, but I hope we still have defaults for everything and an empty CR with env-name will create a single KIE server with default image. More flexibility is good, as long as we maintain simplicity.

I will work on the example CRs. Default functionality should still be the same.

deploy/crds/kieapp.crd.yaml
tchughesiv: what if both are set? does it default to one or the other

deploy/crds/kieapp.crd.yaml
tchughesiv: what if both are picked?

deploy/crds/kieapp.crd.yaml
ruromero: It will fail

tchughesiv: Let’s add some logic to prevent that

ruromero:
> Add logic to prevent deployment failure

It's already there: https://github.com/kiegroup/kie-cloud-operator/pull/63/commits/f46df72ad9e6d10e9dc0834221cee01cab565ce9#diff-53394291f02ee59483694ad21df717b5R114

PR #65, opened 2019-02-15
[KIECLOUD-115] cekit for rhel builds

tchughesiv:
- pushes to dist-git and kicks off build in OSBS

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -1,0 +1,37 @@
+schema_version: 1
+
+name: "rhpam-7-tech-preview/business-automation-operator"
+description: "Red Hat Business Automation Operator"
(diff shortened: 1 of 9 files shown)


image.yaml
```diff
@@ -0,0 +1,37 @@
+schema_version: 1
+
+name: "rhpam-7-tech-preview/business-automation-operator"
+description: "Red Hat Business Automation Operator"
+version: "x.x"
```
ruromero: Can't you just remove this line? Is it perhaps required?

tchughesiv: it's required... just verified.

tchughesiv:
added note in README

PR #68, opened 2019-02-20
[KIECLOUD-128] set default server object in CR

tchughesiv:
defaults to 1 kieserver in CR when neither Server or Servers objects are set.

e.g.
```yaml
objects:
  console:
    resources: {}
  server:
    deployments: 1
    server:
      resources: {}
  smartrouter:
    resources: {}
```

/assign @bmozaffa 

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -131,11 +131,12 @@ func getServersConfig(cr *v1.KieApp, commonConfig *v1.CommonConfig) ([]v1.Server
 			servers = append(servers, crTemplate)
 		}
 	} else {
-		deployments := constants.DefaultKieDeployments
-		if cr.Spec.Objects.Server != nil && cr.Spec.Objects.Server.Deployments != 0 {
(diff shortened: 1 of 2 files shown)


ruromero:
Because if you want to skip the deployment of kieservers you should use `omit: true` otherwise it should always deploy the number defined in constants.

ruromero:
Besides, you can also manually set `server.deployments: 0` and the loop won't deploy any kieserver but as we do for console, if it is not defined it will be deployed.

tchughesiv:
> Because if you want to skip the deployment of kieservers you should use `omit: true` otherwise it should always deploy the number defined in constants.

@ruromero the issue is that it wasn't populating the CR w/ the default number of kieserver deployments. when i view a deployed CR, the server.deployments field is non-existant. we should be updating the CR w/ our default settings so a user can modify those if they choose.

tchughesiv:
> Besides, you can also manually set `server.deployments: 0` and the loop won't deploy any kieserver but as we do for console, if it is not defined it will be deployed.

Yes, i added a test for this scenario as we probably want to keep that ability.

PR #69, opened 2019-02-22
[WIP] Align all labels

ruromero:

As a follow up of #67 

I see more organised the following use of the labels, which is the same pattern I followed in the APB:
* **app**: group the resources of the same type and deployment together. E.g. *rhpam-trial-businesscentral*. This will create a group in the Openshift UI showing the services, dc, secret, pods and route
* **application**: should be the name of the application to be able to virtually group all the resources of the same deployment e.g. *rhpam-trial* will fetch all the service, routes, dc, secrets, pods, ... created for business central, process server, postgres, etc.
* **service**: would retrieve all the resources of any deployment of the same type. e.g. *rhpam-businesscentral*. Identifying routes, dcs, pods, services, etc. from rhpam-trial and rhpam-authoringl but only related to businesscentral.

### DIFF
--- a/config/common.yaml
+++ b/config/common.yaml
@@ -3,9 +3,9 @@ console:
     - metadata:
         name: "{{.ApplicationName}}-{{.Console.Name}}"
         labels:
-          app: "{{.ApplicationName}}"
+          app: "{{.ApplicationName}}-{{.Console.Name}}"
           application: "{{.ApplicationName}}"
-          service: "{{.ApplicationName}}-{{.Console.Name}}"
+          service: "{{.Product}}-{{.Console.Name}}"
       spec:
         strategy:
           type: Recreate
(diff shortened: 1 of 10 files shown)


ruromero:
@tchughesiv that's the whole point of having the 3 labels. With them you could for example:
* Get all the resources of the same kieapp e.g. `application=myapp`
* Get all the resources of the same artifact for a single kieapp e.g. `app=myapp-kieserver-0`
* Get all the resources of the same artifact for any kieapp in the cluster e.g. `service=rhpam-kieserver`

In the UI it will be displayed as follows (multiple kieapps in the same namespace)
- myapp-kieserver-0
- myapp-kieserver-1
- myapp-postgres-0
- myapp-postgres-1
- myapp-rhpamcentr
- otherapp-kiserver-0
- otherapp-rhpamcentr

tchughesiv:
/hold
let's check w/ david and team on this

ruromero:
@errantepiphany what do you think? Shall we stick to what's currently in the templates or do you find this approach useful?

errantepiphany:
@ruromero Can you touch base with Ricardo? We were talking about the app and application labels, if we really need both, grouping by template name, etc. in our last meeting. He was going to come up with recommendation(s) on it. I think you've got some good ideas here, but I would just like to make sure that we end up being aligned (operator with the templates) - even if it takes an iteration or two. Thanks!

tchughesiv:
@ruromero is this something you still wanted to pursue?

ruromero:
This is something I still consider useful but it is so outdated that would have to be redone.
Closing due to lack of interest

PR #72, opened 2019-02-25
[KIECLOUD-131] imageRegistry validation

tchughesiv:

### DIFF
--- a/deploy/examples/image_registry.yaml
+++ b/deploy/examples/image_registry.yaml
@@ -1,0 +1,9 @@
+apiVersion: app.kiegroup.org/v1
+kind: KieApp
+metadata:
+  name: image-registry
+spec:
+  environment: rhpam-trial
+  imageRegistry:
+    insecure: true
+    registry: registry.example.com:5000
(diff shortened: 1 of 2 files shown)


deploy/examples/image_registry.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: app.kiegroup.org/v1
+kind: KieApp
+metadata:
+  name: image-registry
+spec:
+  environment: rhpam-trial
+  imageRegistry:
+    insecure: true
+    registry: registry.example.com:5000
```
bmozaffa: See our validation [unit test,](https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/controller/kieapp/test/crd_validation_test.go#L26). I think we should extend it, or add another function for the examples folder.

tchughesiv: i already added it with a prior PR...
https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/controller/kieapp/test/crd_validation_test.go#L38

PR #85, opened 2019-03-01
[KIECLOUD-52] Consolidate server configurations

tchughesiv:
- moved to a single `servers` array which houses server configs
 - each server config will spawn multiple deploymentConfigs based on its `deployments` field. defaults to 1.
 - default `KieName` is the cr name, but users can specify their own
 - this method called for new `KieName` & `KieIndex` fields in the env server template
 - moved SSO configs under servers object
 - rhel olm bundle file and logic no longer needed

/assign @ruromero @bmozaffa


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -109,50 +106,49 @@ func getConsoleTemplate(cr *v1.KieApp) v1.ConsoleTemplate {
...
+		kieID := GenKieName(cr.Name, cr.Spec.Objects.Servers[index].Name)
+		if exists := names[kieID]; exists {
+			return []v1.ServerTemplate{}, fmt.Errorf("duplicate kieserver name %s", kieID)
 		}
-	} else {
-		if cr.Spec.Objects.Server == nil {
-			cr.Spec.Objects.Server = &v1.CommonKieServerSet{
-				Deployments: constants.DefaultKieDeployments,
-			}
+		if cr.Spec.Objects.Servers[index].Deployments == nil {
```
bmozaffa: @ruromero , why do we need to support "deployments = 0"?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -109,50 +106,49 @@ func getConsoleTemplate(cr *v1.KieApp) v1.ConsoleTemplate {
...
-			kieServerID := fmt.Sprintf(defaultKieServerIDTemplate, cr.Name, i)
-			if len(server.Name) > 0 {
-				kieServerID = server.Name
-			}
-			crTemplate := v1.ServerTemplate{
-				KieServerID: kieServerID,
-			}
-			crTemplate.Build = getBuildConfig(commonConfig, server.Build)
-			if server.Build != nil {
-				crTemplate.From = getKieServerImageForBuild(commonConfig, i)
```
bmozaffa: @ruromero , we decided it doesn't make sense to have deployments > 1, when it's a build, right? Am I then correct to assume this should not have been there?

PR #100, opened 2019-03-15
[KIECLOUD-149] Fix trial environments

ruromero:

Fix https://issues.jboss.org/browse/KIECLOUD-149

* Remove h2 configuration as it is already the default one
* Remove ping service and env vars for rhpamcentr in trial environments
* Remove kie_server_port and protocol env vars
* Use podIP as KIE_SERVER_HOST also for trial envs
* Trial server replicas set to 1

### DIFF
--- a/pkg/controller/kieapp/defaults/merge_test.go
+++ b/pkg/controller/kieapp/defaults/merge_test.go
@@ -46,19 +46,19 @@ func TestMergeServices(t *testing.T) {
 	overwrite.Console.Services = append(overwrite.Console.Services, *service5)
 
 	mergedEnv, _ := merge(baseline, *overwrite)
-	assert.Equal(t, 4, len(mergedEnv.Console.Services), "Expected 4 services")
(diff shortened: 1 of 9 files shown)


pkg/controller/kieapp/defaults/merge_test.go
```diff
@@ -46,19 +46,19 @@ func TestMergeServices(t *testing.T) {
 	overwrite.Console.Services = append(overwrite.Console.Services, *service5)
 
 	mergedEnv, _ := merge(baseline, *overwrite)
-	assert.Equal(t, 4, len(mergedEnv.Console.Services), "Expected 4 services")
+	assert.Equal(t, 3, len(mergedEnv.Console.Services), "Expected 4 services")
```
tchughesiv: should be "Expected 3 services"

ruromero:
https://github.com/kiegroup/kie-cloud-operator/blob/master/config/common.yaml#L493-L496
But was being overridden here only for the trial environment
https://github.com/kiegroup/kie-cloud-operator/blob/master/config/envs/rhdm-trial.yaml#L66

PR #102, opened 2019-03-18
CommonConfig clean up

ruromero:
* Move auth, mavenRepo and Product out of commonConfig
* Remove the same attributes from CRD
* Rename DB examples
* Added new object `templateConstants` to limit the use of fixed literals


pkg/apis/app/v1/kieapp_types.go
```diff
@@ -359,17 +366,16 @@ type BuildTemplate struct {
...
-	MavenPassword      string       `json:"mavenPassword,omitempty"`
+	ApplicationName    string `json:"applicationName,omitempty"`
+	Version            string `json:"version,omitempty"`
+	ImageTag           string `json:"imageTag,omitempty"`
+	Product            string `json:"product,omitempty"`
+	KeyStorePassword   string `json:"keyStorePassword,omitempty"`
+	AdminPassword      string `json:"adminPassword,omitempty"`
+	ControllerPassword string `json:"controllerPassword,omitempty"`
+	ServerPassword     string `json:"serverPassword,omitempty"`
+	MavenRepo          string `json:"mavenRepo,omitempty"`
```
tchughesiv: should this be moved to template constants? elsewhere?

ruromero: I think we should leave this credentials configurable

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -351,6 +353,29 @@ func (reconciler *Reconciler) dcUpdateCheck(current, new oappsv1.DeploymentConfi
...
+func (reconciler *Reconciler) hasUpdatedVolumes(current, new corev1.PodSpec) bool {
+	cContainer := current.Containers[0]
+	nContainer := new.Containers[0]
+
+	if len(current.Volumes) != len(new.Volumes) || len(cContainer.VolumeMounts) != len(nContainer.VolumeMounts) {
+		return true
+	}
+
+	for i := range current.Volumes {
+		if !strings.HasSuffix(current.Volumes[i].Name, constants.KeystoreVolumeSuffix) && !reflect.DeepEqual(current.Volumes[i], new.Volumes[i]) {
```
tchughesiv: can we expect that the ordering of old and new volume/mount arrays will be identical? should we sort them by name first to ensure proper comparison? i'm not aware of whether ocp does any ordering post deployment.

pkg/apis/app/v1/kieapp_types.go
tchughesiv: agree, what about  `MavenRepo` though?

pkg/controller/kieapp/kieapp_controller.go
ruromero: Removed this logic

PR #106, opened 2019-03-26
[KIECLOUD-168] rhpam authoring-ha fixes

tchughesiv:

### DIFF
--- a/config/envs/rhpam-authoring-ha.yaml
+++ b/config/envs/rhpam-authoring-ha.yaml
@@ -58,7 +58,119 @@ smartRouter:
 ## KIE smartrouter END
 # ES/AMQ BEGIN
 others:
-  - persistentVolumeClaims:
+  - roles:
+      - metadata:
+          name: "[[.ApplicationName]]-amq-role"
+          labels:
+            app: "[[.ApplicationName]]"
+            application: "[[.ApplicationName]]"
+        rules:
+          - apiGroups:
+              - ""
+            resources:
+              - endpoints
+            verbs:
+              - create
+              - delete
+              - deletecollection
+              - get
+              - list
+              - patch
+              - update
+              - watch
+          - apiGroups:
+              - ""
+            resources:
+              - namespaces
+            verbs:
+              - get
+              - list
+
+      - metadata:
+          name: "[[.ApplicationName]]-amq-scaledown-controller-role"
+          labels:
+            app: "[[.ApplicationName]]"
+            application: "[[.ApplicationName]]"
+        rules:
+          - apiGroups:
+              - apps
+            resources:
+              - statefulsets
+            verbs:
+              - get
+              - list
+              - watch
+          - apiGroups:
+              - ""
+            resources:
+              - pods
+            verbs:
+              - get
+              - list
+              - watch
+              - create
+              - delete
+              - patch
+              - update
+          - apiGroups:
+              - ""
+            resources:
+              - persistentvolumeclaims
+            verbs:
+              - get
+              - list
+              - watch
+              - delete
+          - apiGroups:
+              - ""
+            resources:
+              - events
+            verbs:
+              - get
+              - list
+              - watch
+
+    rolebindings:
+      - metadata:
+          name: "[[.ApplicationName]]-amq-scaledown-controller-openshift-rb"
+          labels:
+            app: "[[.ApplicationName]]"
+            application: "[[.ApplicationName]]"
+        subjects:
+          - kind: ServiceAccount
+            name: "[[.ApplicationName]]-amq-scaledown-controller-sa"
+        roleRef:
+          kind: Role
+          name: "[[.ApplicationName]]-amq-scaledown-controller-role"
+          apiGroup: rbac.authorization.k8s.io
+
+      - metadata:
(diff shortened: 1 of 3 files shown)


config/envs/rhpam-authoring-ha.yaml
```diff
@@ -138,17 +250,172 @@ others:
...
+              terminationGracePeriodSeconds: 30
+          triggers:
+            - imageChangeParams:
+                automatic: true
+                containerNames:
+                  - "[[.ApplicationName]]-amq"
+                from:
+                  kind: ImageStreamTag
+                  namespace: "openshift"
+                  name: "amq-broker-72-scaledown-controller-openshift:0.4"
```
sutaakar: Why there is a specified version 0.4 here when template is using 1.0? https://github.com/jboss-container-images/rhpam-7-openshift-image/blob/7.3.x/templates/rhpam73-authoring-ha.yaml#L288

The 1.0 version from template is mapped to 0.4 in AMQ image stream.

tchughesiv: @sutaakar where in that template is the imagestream mapping to 0.4 created?

sutaakar: @tchughesiv You can find it in https://raw.githubusercontent.com/jboss-container-images/jboss-amq-7-broker-openshift-image/amq-broker-72/amq-broker-7-scaledown-controller-image-streams.yaml

tchughesiv: @sutaakar ok, but are we saying this imagestream will be available in next 3.11 release? 4.0?

sutaakar: I am not sure when it becomes available. Just pointing how is the image stream defined in our templates and how it is mapped in AMQ image stream file.

tchughesiv: i get it... but the issue we have here is, what if using the operator in a cluster that doesn't have those AMQ imagestreams? the DC will attempt to pull amq-broker-72-scaledown-controller-openshift:1.0 from our registry. That image doesn't exist so the env will never deploy. Additionally, the operator has the ability to create missing imagestreams for the user... but if it tries to create one to a non-existent image, that's no use. 

I can certainly change it to 1.0 but not sure it makes much sense to.

sutaakar: hmm, I see. In that case I would leave it as it is now, can be adjusted later.

errantepiphany: There is obviously a difference between an image _stream_ version and the _image_ version/tag that it references. In this case, as you noted, the AMQ project's image streams file defines a 1.0 stream that points to a 0.4 image:
https://github.com/jboss-container-images/jboss-amq-7-broker-openshift-image/blob/amq-broker-72/amq-broker-7-scaledown-controller-image-streams.yaml#L26-L29
Personally, I wish they were the same. Specifically, I wish they were both 0.4, so that they can both become 1.0 at the same time. However, it's not my project...
Maybe the best thing to do is see how the stream is versioned in the OCP catalog? Or ask the maintainers of that project?

tchughesiv: i see your point too though... let's just let @errantepiphany make the call :)

errantepiphany: https://github.com/jboss-container-images/jboss-amq-7-broker-openshift-image/graphs/contributors

tchughesiv: @sutaakar tks for calling this out... i've made the switch to `1.0` as you suggested.

tchughesiv:
@bmozaffa @sutaakar we should be good to go now

config/envs/rhpam-authoring-ha.yaml
```diff
@@ -228,6 +529,23 @@ others:
...
+      - spec:
+          clusterIP: None
+          ports:
+            - port: 8888
+              protocol: TCP
+              targetPort: 8888
+          selector:
+            deploymentConfig: "[[.ApplicationName]]-amq"
+        metadata:
+          name: "[[.ApplicationName]]-amq-ping"
```
sutaakar: I suppose that this is AMQ scaledown controller ping service. The name of this service needs to be "ping" (as used in template https://github.com/jboss-container-images/rhpam-7-openshift-image/blob/7.3.x/templates/rhpam73-authoring-ha.yaml#L770). The reason is that ping service for AMQ scaledown controller is hardcoded, see AMQ scaledown controller

tchughesiv: @sutaakar 
i set this env var to resolve that issue and allow for custom service names:
https://github.com/kiegroup/kie-cloud-operator/blob/f308ce8d29101824a90b75ed20d2b7d797a144d9/config/envs/rhpam-authoring-ha.yaml#L452-L453

i get the following in logs and amq runs fine:
```shell
2019-03-29 13:58:26,630 INFO  [org.jgroups.protocols.openshift.DNS_PING] serviceName [rhpam-authoring-ha-amq-ping] set; clustering enabled
```

sutaakar:
Will check the functionality once merged and available in image.

PR #114, opened 2019-04-05
check for openshift cluster before watching objects

tchughesiv:
/assign @bmozaffa @sutaakar 


bmozaffa:
@sutaakar FYI, this is part of an attempt to reuse common operator code across all of middleware (if not more). The starting point here, which is fairly trivial, is that we don't run on pure k8s and many other operators also need to know if running on ocp or k8s

sutaakar:
ok

PR #120, opened 2019-04-09
fix rhel build

spolti:

### DIFF
--- a/modules/kie-cloud-operator/install
+++ b/modules/kie-cloud-operator/install
@@ -1,8 +1,6 @@
 #!/bin/sh
 
 INSTALL_PKGS="go-toolset-1.10"
-yum install -y ${INSTALL_PKGS} --setopt=tsflags=nodocs \
-    && rpm -q ${INSTALL_PKGS}
 
 scl enable go-toolset-1.10 - << \EOF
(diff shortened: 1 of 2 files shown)


modules/kie-cloud-operator/install
```diff
@@ -1,9 +1,5 @@
 #!/bin/sh
 
-INSTALL_PKGS="go-toolset-1.10"
```
tchughesiv: This var is used to remove go later in the script.

spolti: Sry, missed that.

Yes, the image size is under 400mb:
2019-04-10 09:51:11,496 root         INFO     Original image size: 922.69 MB
2019-04-10 09:51:11,496 root         INFO     Squashed image size: 376.81 MB


Just to be sure, when squashing do you select a layer or squash all layers?

tchughesiv: whatever is default in brew

tchughesiv:
@spolti looks like these changes will increase our image size a bit.

currently we're at 394 MB
w/ these changes we bump up to 421 MB

I think its worth the extra 27mb to merge these changes.
/lgtm

PR #133, opened 2019-04-17
[JBPM-8150] Allow users to change Kie server id in Deployment config

tchughesiv:
/assign @bmozaffa @rhtevan

 - still not defaulting to OpenShiftStartupStrategy, but include an example that enables it
 - adds ability to set server id within CR spec
 - changes appropriate server ConfigMap label to "DETACHED" when the operator detects a server's replicas set to zero

### DIFF
--- a/pkg/controller/kieapp/kieapp_controller.go
+++ b/pkg/controller/kieapp/kieapp_controller.go
@@ -82,7 +82,6 @@ func (reconciler *Reconciler) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	dcUpdated, err := reconciler.updateDeploymentConfigs(instance, env)
-
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -96,6 +95,9 @@ func (reconciler *Reconciler) Reconcile(request reconcile.Request) (reconcile.Re
 	if bcUpdated && status.SetProvisioning(instance) {
 		return reconciler.UpdateObj(instance)
 	}
+
+	// Check the KieServer ConfigMaps for necessary changes
+	reconciler.checkKieServerConfigMap(instance, env)
 
 	// Fetch the cached KieApp instance
 	cachedInstance := &v1.KieApp{}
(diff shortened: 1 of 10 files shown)


bmozaffa:
@rhtevan I'll wait for your ack

tchughesiv:
@bmozaffa @rhtevan its ready now

bmozaffa:
Great, thanks, will wait for @rhtevan first

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -769,3 +777,89 @@ func (reconciler *Reconciler) createConfigMap(obj v1.OpenShiftObject) (*corev1.C
...
+					// Create request
+					req, err := http.NewRequest(method, url, nil)
+					if err != nil {
+						log.Error(err)
+					}
+
+					// Fetch Request
+					req.Header.Set("Content-Type", "application/json")
+					req.SetBasicAuth(instance.Spec.CommonConfig.AdminUser, instance.Spec.CommonConfig.AdminPassword)
+					resp, err := client.Do(req)
```
bmozaffa: @rhtevan this goes against our dependency hierarchy. The operator is an infrastructure component that only interacts with Kubernetes resources and the k8s API. BC should watch or poll the kube API as needed.

tchughesiv: @rhtevan i strongly agree and the more i test, i'm finding certain scenarios are crashing the operator. i think i'd be ok with the labels logic if absolutely necessary, but i'd like to at least remove the BC api logic from this PR.

rhtevan:
@bmozaffa @tchughesiv 
Based on my limited knowledge of GoLang and Operator SDK, it looks fine with me, except that it would be nice to do the http 'ping' against individual BC/WB Pod endpoints instead of against BC/WB Service endpoint. Having said that, if it is too much hassle to do it I'm fine for now.

pkg/controller/kieapp/kieapp_controller.go
rhtevan: @tchughesiv 
If it is the case, then we should take the BC logic out. Please keep the labeling logic. I'll explore alternatives to tackle.

tchughesiv:
@bmozaffa @rhtevan i've removed the BC requests, but left the labeling logic.

PR #134, opened 2019-04-18
[KIECLOUD-189] Updated for operator-sdk 0.7

bmozaffa:


Gopkg.toml
```diff
@@ -47,8 +64,8 @@ required = [
 [[constraint]]
   name = "github.com/operator-framework/operator-sdk"
   # The version rule is used for a specific release and the master branch for in between releases.
-  # branch = "master" #osdk_branch_annotation
-  version = "=v0.3.0" #osdk_version_annotation
+  branch = "v0.7.x" #osdk_branch_annotation
```
tchughesiv: should this be v0.7.0 instead?

bmozaffa: You're right, I wasn't going off the tag. Fixed and being pushed now.

PR #136, opened 2019-04-18
[KIECLOUD-190] KIE Server needs to respect configured imageTag

bmozaffa:
KIE Server should also use any provided commonConfig imageTag, as console and smart router do


bmozaffa:
@tchughesiv this just needs a rebase and should be a very small PR on top of previous one

PR #138, opened 2019-04-18
[KIECLOUD-190] KIE Server needs to respect configured imageTag

bmozaffa:
Cherry-pick from master
KIE Server should also use any provided commonConfig imageTag, as console and smart router do

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -413,7 +413,7 @@ func getDefaultKieServerImage(product string, config *v1.CommonConfig, from *cor
 	if from != nil {
 		return *from
 	}
-	imageName := fmt.Sprintf("%s%s-kieserver-openshift:%s", product, config.Version, constants.ImageStreamTag)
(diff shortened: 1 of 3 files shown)


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -413,7 +413,7 @@ func getDefaultKieServerImage(product string, config *v1.CommonConfig, from *cor
 	if from != nil {
 		return *from
 	}
-	imageName := fmt.Sprintf("%s%s-kieserver-openshift:%s", product, config.Version, constants.ImageStreamTag)
+	imageName := fmt.Sprintf("%s%s-kieserver-openshift:%s", product, config.Version, config.ImageTag)
```
tchughesiv: @bmozaffa can we limit this PR against 1.0.x branch to only this change? i'm thinking we should avoid an sdk upgrade in that branch as its only for the 7.3.x releases.

bmozaffa: Other changes removed, ready for review

PR #148, opened 2019-04-22
[KIECLOUD-194] Deployment podStatuses, edge case bug fix

bmozaffa:
Adding generated functions for new status structs


sutaakar:
@bmozaffa Can you please add test coverage for the change?

bmozaffa:
Hi @sutaakar , Tommy and I separately tested this quite a bit in our lab, but it is very difficult to capture in unit tests. It's a question of reconciling objects back and forth with Kubernetes, and it either needs a k8s environment as part of the test suite, or a good amount work to mock one. Thoughts?

sutaakar:
In Kie server test suite Evan used mocked OpenShift server for testing purposes (based on Java). Is there something similar available in Go?

bmozaffa:
Yes, there is a mock client, but it doesn't do anything. We would have to write enough logic to react to operations, for example maintain a cache of objects so when a get() is called, we can return them, to add to them with create(), and to update them as appropriate. In a case like this, we'd need to do even more... we need to use a timer to simulate gradual transition of the server from starting to ready and so on. It is doable, just not quick or easy! :-(

sutaakar:
hmm, I see. So this would be better to cover by integration tests against live OpenShift instance, right?

bmozaffa:
Yes, much easier to cover in integration tests

bmozaffa:
@sutaakar I was right that this wasn't even to cover in unit tests! :-)

But it is done now: https://github.com/kiegroup/kie-cloud-operator/pull/155

PR #153, opened 2019-04-26
[KIECLOUD-136] - Add the RHPAM/DM JMS support to operator

spolti:


deploy/crds/kieapp.1.0.crd.yaml
```diff
@@ -199,7 +199,7 @@ spec:
                         properties:
                           kieServerContainerDeployment:
                             type: string
-                            description: The Maven GAV to deploy, e.g., rhpam-kieserver-library=org.openshift.quickstarts:rhpam-kieserver-library:1.4.0-SNAPSHOT
+                            description: The Maven GAV to deploy, e.g., rhpam-kieserver-library=org.openshift.quickstarts:rhpam-kieserver-library:1.5.0-SNAPSHOT
```
tchughesiv: Changing this file is unnecessary as it needs to match the codebase in our 1.0.x branch. I'm not entirely sure we'll keep archiving crd's this way long term.

spolti: ok, I only updated that because the quickstart version was updated on master branch, also update the crd to keep the quickstart versions aligned. Should I undo it?

tchughesiv: quickstart version?

bmozaffa: To be clear, as Tommy is saying, this CRD is for our 7.3 version only, whereas the other one is for 7.4. @spolti, should this change retroactively apply to previous 7.3 releases, or at least to the next 7.3 patch?

spolti: Yes, the quickstart used with s2i builds, for the immutable environments.
https://github.com/jboss-container-images/rhpam-7-openshift-image/tree/master/quickstarts/library-process/library

spolti: @bmozaffa I think so, otherwise the immutable cr provided as example will not work without change  the quickstart version.

But, thinking better, would probably better use the tag version for the quickstart instead master branch.

config/jms/activemq-jms-config.yaml
```diff
@@ -0,0 +1,225 @@
...
+                    value: "[[.Jms.KieServerJmsAMQQueues]]"
+                  # JMS config END
+    ## KIE server deployment config END
+    ## AMQ deployment BEGIN
+    - metadata:
+        name: "[[.KieName]]-amq"
+        labels:
+          app: "[[$.ApplicationName]]"
+          application: "[[$.ApplicationName]]"
+          service: "[[.KieName]]-kieserver"
```
tchughesiv: This should probably just be:
`          service: "[[.KieName]]"`

like we see here in other config files:
https://github.com/kiegroup/kie-cloud-operator/blob/9547c5f534eb7a09e6dea1531cfa7f0f01bdfa74/config/common.yaml#L433

tchughesiv: unless you intended this to be:
`service: "[[.KieName]]-amq"` ??

spolti: that's right, it should be .KieName

PR #154, opened 2019-04-28
[KIECLOUD-177] Operator UI integration

bmozaffa:
This PR provides an initial integration with the console-cr-form project to provide a UI for operator custom resources. This includes building and deploying the UI, but so far stops short of accepting new or updated yaml from the UI.


sutaakar:
@bmozaffa Is there any image which QE can use to check the functionality from this PR?

bmozaffa:
Hi @sutaakar, yes, it's at quay.io/bmozaffa/kie-cloud-operator:1.1 and currently deployed in our lab, so the UI itself is also accessible at http://console-cr-form-operator-ui.app.middleware.ocp.cloud.lab.eng.bos.redhat.com/

sutaakar:
Thanks

pkg/controller/initializer.go
```diff
@@ -13,6 +13,7 @@ func init() {
 	addManager := func(mgr manager.Manager) error {
 		k8sService := GetInstance(mgr)
 		reconciler := kieapp.Reconciler{Service: &k8sService}
+		deployConsole(&reconciler)
```
tchughesiv: I'm thinking we only really want this to deploy in the event we're running the operator as a deployment in openshift. We already do something similar for creating the env ConfigMaps here:
https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/controller/kieapp/kieapp_controller.go#L47

The other big benefit of moving this method call to that function instead is we can set an ownerRef of that main operator deployment on the console's pod/service/route. Thoughts? 

I expect we could also do away with the IMAGE var but just parsing the deployment object and grabbing the image setting dynamically.

bmozaffa: @tchughesiv These changes are now pushed

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -53,6 +53,7 @@ func (reconciler *Reconciler) Reconcile(request reconcile.Request) (reconcile.Re
 		} else {
 			log.Error("Can't properly create ConfigMaps. ", err)
 		}
+		deployConsole(reconciler, myDep)
```
tchughesiv: this should probably be moved up under line 52 (reconciler.CreateConfigMaps) ... need to be sure `myDep` was successfully pulled first.

pkg/controller/kieapp/deploy_ui.go
```diff
@@ -0,0 +1,167 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"os"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+var name = "console-cr-form"
+var app = "kie-cloud-operator"
```
tchughesiv: better to use `opName` which is set and passed back from `UseEmbeddedFiles` as seen here:
https://github.com/kiegroup/kie-cloud-operator/blob/6478199f839626fd9a68bba11431b5e5a992cb8f/pkg/controller/kieapp/defaults/defaults.go#L582

this can be passed into `deployConsole` method as new input instead. This is preferable because our product release name is actually business-automation-operator but its set dynamically.

PR #165, opened 2019-05-14
Console update

bmozaffa:
Updated to use the latest console-cr-form
Integration on operator side to provide form metadata to console
Unit tests for form/schema/etc


pkg/ui/serve.go
```diff
@@ -132,11 +96,28 @@ func getSchema() spec.Schema {
 	crd := &CustomResourceDefinition{}
 	err = yaml.Unmarshal(yamlByte, crd)
 	if err != nil {
+		log.Fatal(err)
 		panic("Failed to unmarshal static schema, there must be an environment problem!")
 	}
 	return crd.Spec.Validation.OpenAPIV3Schema
 }
 
+func getForm() web.Form {
```
ruromero: I don't think it was strictly necessary but I prefer to have it in a separate file

bmozaffa: btw built version is at https://console-cr-form-operator-ui.app.middleware.ocp.cloud.lab.eng.bos.redhat.com

PR #170, opened 2019-05-18
align with latest console-ui

ruromero:
Sync with latest changes from console project


deploy/ui/form.json
```diff
@@ -120,7 +141,6 @@
                 {
                   "label": "url",
                   "type": "text",
-                  "required": true,
```
bmozaffa: Why are these 3 set to not being required any longer?

deploy/ui/form.json
```diff
@@ -19,33 +19,56 @@
           "originalJsonPath": "$.spec.environment",
           "type": "dropDown"
         },
+
         {
-          "label": "Image Registry",
-          "default": "rhpam-trial",
+          "label": "External registry",
```
bmozaffa: Is there a technical restriction to using the same label of "Image Registry"?

ruromero: No, there isn't

deploy/ui/form.json
ruromero: Restored

PR #173, opened 2019-05-21
[KIECLOUD-107] - Add flag to disable prometheus ext

ricardozanini:
Please see:
https://issues.jboss.org/browse/KIECLOUD-107

**Opening for discussion**

**Use Case:**
The user should choose if the Prometheus Server Extension will be disabled or not. Added a dropdown to the UI which will able the user to choose whether to disabled it.

If not set, the var `PROMETHEUS_SERVER_EXT_DISABLED` **should not be** added to the kieserver, otherwise, follow the what was set by user.

This could be handled just in the env variables, but a field in the interface would improve UX.

Signed-off-by: Ricardo Zanini <zanini@redhat.com>


ricardozanini:
@ruromero some tests from the master branch are not running in my local env. I'd like to add more tests cases to test this scenario, should reach you out tomorrow morning.

pkg/apis/app/v1/kieapp_types.go
```diff
@@ -108,16 +108,29 @@ type KieAppObjects struct {
...
-	Jms                 *KieAppJmsObject `json:"jms,omitempty"`
+	Deployments          *int                    `json:"deployments,omitempty"` // Number of KieServer DeploymentConfigs (defaults to 1)
+	Name                 string                  `json:"name,omitempty"`
+	ID                   string                  `json:"id,omitempty"`
+	From                 *corev1.ObjectReference `json:"from,omitempty"`
+	Build                *KieAppBuildObject      `json:"build,omitempty"` // S2I Build configuration
+	SecuredKieAppObject  `json:",inline"`
+	Database             *DatabaseObject                 `json:"database,omitempty"`
+	Jms                  *KieAppJmsObject                `json:"jms,omitempty"`
+	DisablePrometheusExt PrometheusExtDisabledChoiceType `json:"disablePrometheusExt,omitempty"`
```
bmozaffa: Do we really need 3 modes for this (i.e., enabled, disabled, and unspecified)?

If we don't need three modes, let's make it a bool that is disabled by default, unless enabled.

If we need three modes, create a struct for it that has an enabled as a bool

pkg/apis/app/v1/kieapp_types.go
```diff
@@ -108,16 +108,29 @@ type KieAppObjects struct {
...
+type PrometheusExtDisabledChoiceType string
+
+const (
+	// NotDefined if user didn't set the option
+	NotDefined PrometheusExtDisabledChoiceType = ""
+	// Enabled extension should be enabled
+	Enabled PrometheusExtDisabledChoiceType = "Enabled"
+	// Disabled extension should be disabled
+	Disabled PrometheusExtDisabledChoiceType = "Disabled"
+)
```
bmozaffa: This would go away based on the struct, see below.

deploy/crds/kieapp.crd.yaml
```diff
@@ -438,6 +438,13 @@ spec:
                           kieServerJmsAMQQueues:
                             type: string
                             description: AMQ broker broker comma separated queues, if empty the values from default queues will be used.
+                disablePrometheusExt:
+                  type: string
+                  description: An option indicating if prometheus server extension should be disabled
+                  enum:
+                    -
+                    - Enabled
+                    - Disabled
```
bmozaffa: Should be bool, possibly in a struct that's a pointer, see comment under kieapp_types

pkg/apis/app/v1/kieapp_types.go
ricardozanini: The original design was a checkbox. But the requirement has 3 states: undefined, true or false. If undefined the env variable shouldn't exist. That's why I opt for adding a dropdown instead a checkbox. But, I have to agree that this makes little sense.

@errantepiphany could you clarify? Don't you think that this requirement is a bit odd?

> The way the templates and code are now, the system property is only set IF a value for the PROMETHEUS_SERVER_EXT_DISABLED parameter is set. We do not default it to anything.

It's a boolean value, so defaulting to FALSE (unchecked), would make sense for this scenario since the prometheus extension would be enabled by default. Doing this in templates would be different from the APB and Operator since they have UI.

errantepiphany: @ricardozanini I'm okay with the checkbox, and defaulting to unchecked (disabled = false) value.

bmozaffa:
Another comment, @ricardozanini and @errantepiphany , just to make sure we're on the same page. Normally, environment variables can simply be added to the CR under the Env section. See [here](https://github.com/kiegroup/kie-cloud-operator/blob/master/deploy/examples/server_config.yaml#L10) for an example. That means this PR would not be necessarily at all for disabling prometheus. The only question is how important this flag is and to what extent we want to expose it. At some point, we will want to stop adding items, because if we have 20 configuration options, the whole argument of making it more prominent by including it in the top-level CR no longer makes sense. Thoughts?

bmozaffa:
Please also see failed unit tests!

ricardozanini:
@bmozaffa I'm sorry for the failed unit tests, I pushed by mistake to discuss the implementation with Spolti. This commit wasn't ready yet.

About your concern, I fell you have a point since there's a lot of fields in the UI and this could be too much. Prometheus monitoring seems very important for the end users and have this option OOTB would make their life easier. But it's only one option. Maybe leaving in the docs as an optional ENV var and defaulting it to `false`, which enable the prometheus integration, could be enough. This decision is not mine, though. @errantepiphany thoughts?

errantepiphany:
@ricardozanini , @bmozaffa : I can see the very real possibility of configuration bloat.Generally, whenever we add something to the templates, we try to add it in the APB and Operator too, per previous direction. However, maybe we should decide on inclusion of new parameters on a case-by-case basis. I will default to Babak's judgement on this. I'd also like to hear @spolti 's thoughts.

ricardozanini:
@errantepiphany @spolti @bmozaffa in case you guys decided to add the config, I just pushed a version that add a new type: `prometheus` foreseeing more prometheus related configuration in the future (@spolti idea :+1: ). There are changes in the UI also by adding a checkbox, defaulting to `false`.

bmozaffa:
Thanks, guys, will wait to hear from @spolti as well. It's a broad usability issue and not a specific operator one, so whatever makes sense, works for me. However, with template, you need a template parameter to support configuration. Here, you don't. You can add any arbitrary env variable directly into the CR for the VMs you want. We should certainly decide case by case.

deploy/crds/kieapp.crd.yaml
```diff
@@ -438,6 +438,13 @@ spec:
                           kieServerJmsAMQQueues:
                             type: string
                             description: AMQ broker broker comma separated queues, if empty the values from default queues will be used.
+                      prometheus:
+                        type: object
+                        description: Configuration for Prometheus Server with KIE Server
+                        properties:
+                          disablePrometheusExt:
+                            type: boolean
+                            description: If set to false, the prometheus server extension will be enabled. (Sets the org.kie.prometheus.server.ext.disabled system property)
```
bmozaffa: Sorry to be a pain, but if we do want to do this and don't care about having 3 states (undefined), then no need for a prometheus object, unless you expect further configuration in the future that you want to group under it. Otherwise, the disablePrometheusExt directly in there would work.

spolti:
I suggested that way because, probably new prometheus configurations will appear in the feature (not 100% sure of that), with that said, it is a nice to have something already done on operator where you can only add new fields.

But if this config is just as it is, a single one, so, I'd agree with Babak about just add the env on the CR.

bmozaffa:
Ack, @spolti , now the biggest question is what should be bubbled up in the CR and what should simply be an env key/value pair? If we want to single out 10 configuration items (our of possible env vars) to expose to end users in the long term, including items that do not exist today, do you think this would be one of them?

ricardozanini:
**Thinking in the long term**, I'd say that if we can afford to put more effort to UX, every property should be add to it. Hence the customers would have a nice way to customize their deployment without having to look after the docs for an env variable that will enable/set that hidden feature for him. The UI could "ask" to the customer which use case they'd like to implement and/or what components they are interested in customizing. I know that I'm dreaming and inventing use cases, but as a customer I'd like to see something like this. 

The "env" object in the Operator I'd left for custom env variables (like the JAVA_DEBUG param for example) that the customer need to add to their deployment, rather then use it to add our own env vars, like in this example.

errantepiphany:
> I suggested that way because, probably new prometheus configurations will appear in the feature (not 100% sure of that), with that said, it is nice to have something already done on operator where you can only add new fields when needed.

What if it wasn't a "prometheus" object, but a "metrics" object? That way it's a bit more generic and targets the requirement of metrics, and not a specific implementation or tool. But we could add prometheus config points to the "metrics" object, as well as any future metrics-y stuff.

bmozaffa:
> What if it wasn't a "prometheus" object, but a "metrics" object? That way it's a bit more generic and targets the requirement of metrics, and not a specific implementation or tool. But we could add prometheus config points to the "metrics" object, as well as any future metrics-y stuff.

I think the question is more how common this use case is. We can try to anticipate what the next 10 config items are and if there will be metrics items, but really, who knows? :-)

sutaakar:
Just one comment, considering that Operator will become supported we should also focus on keeping backward compatibility for future versions. Structure changes in CR between minor versions are not ideal.

errantepiphany:
@bmozaffa @ricardozanini @spolti @sutaakar So how do we come to a decision on this? Babak, do you feel strong that it "should simply be an env key/value pair?" @tchughesiv , care to chime in?

ricardozanini:
> Structure changes in CR between minor versions are not ideal.

Good point. If we'll keep compatibility with the tech preview version, the key/value pair should do. Then  add to the docs the prometheus disabled flag env var.

> What if it wasn't a "prometheus" object, but a "metrics" object? That way it's a bit more generic and targets the requirement of metrics, and not a specific implementation or tool. But we could add prometheus config points to the "metrics" object, as well as any future metrics-y stuff.

I'd say for this UC specifically to let it as "prometheus" and thinking about the "generic" solution once it comes up.

spolti:
Yeah, I agree with @sutaakar regarding minor changes. So far the prometheus has no configurations on our side besides this one, it may have a few in the future when the metrics collections for kieserver are done, a input from Vladimir Blagojevic would be helpful to decide. 

Also if we decide to use a generic struct type, then I agree with @errantepiphany to use 'metrics' instead 'prometheus'.

@bmozaffa is right, for this specific use case a env should be enough.

bmozaffa:
> @bmozaffa @ricardozanini @spolti @sutaakar So how do we come to a decision on this? Babak, do you feel strong that it "should simply be an env key/value pair?" @tchughesiv , care to chime in?

I don't feel strongly about it. I just hoped a team discussion would result in a decision and common understanding going forward.

ricardozanini:
@bmozaffa @errantepiphany once you guys decide if this PR will be merged, please let me know to change the default value to `TRUE`.

bmozaffa:
> @bmozaffa @errantepiphany once you guys decide if this PR will be merged, please let me know to change the default value to `TRUE`.

To sum up, I have no strong opinions on the subject, and feel like I lack enough information about how often customers will need to change default behavior, to judge if this merits being bubbled up to the top or not.

@spolti thinks it's unnecessary, and @sutaakar warns about CRD changes in minor versions. Nothing yet from @tchughesiv or @ruromero ?

errantepiphany:
I keep waffling on this. I _do_ like the idea of having complete config option coverage, _but_ given that:
- it's a single, lonely metrics parameter all by itself, so having a separate config object for it seems unnecessary at this point, but including it in the core config also doesn't feel like the right place
- "disabled by default" could actually be considered a "good" thing since it's more lightweight
- enabling it via the env config is very easy to do (as would be the docs instructions)
- the UI isn't advanced enough yet to ask and guide the user through different use cases
, then perhaps we don't include it at this time.

ruromero:
If I can still give my opinion... To me it doesn't seem to be something that users will be changing quite often so I feel better to not update the CRD just for this. If users want to change the value they can do it either adding the corresponding ENV VAR or adding it to the Env ConfigMaps.
Later on if this option becomes popular and we hear that users are setting it as true quite often we just need to bubble it up. However if we decide we want it to be TRUE by default later on, nothing changes just the default value.

tchughesiv:
> @bmozaffa @ricardozanini @spolti @sutaakar So how do we come to a decision on this? Babak, do you feel strong that it "should simply be an env key/value pair?" @tchughesiv , care to chime in?

My vote is, we just add it as an env var to the common.yaml set to `false` as a default. We can even add an example CR showing how to disable it. In the future, if the metrics config piece gets very complex and warrants a field in the CR spec, we can address it then.

ricardozanini:
@tchughesiv @bmozaffa @spolti @errantepiphany I think we can safely close this PR and not merge it then. One day, when metrics would be a thing, I can use it as reference. The doc team will add the instructions of how to use this feature (by setting the env var) in the documentation.

bmozaffa:
> My vote is, we just add it as an env var to the common.yaml set to `false` as a default. We can even add an example CR showing how to disable it. In the future, if the metrics config piece gets very complex and warrants a field in the CR spec, we can address it then.

If it defaults to false, why even add it to common.yaml? Agree on example.

ricardozanini:
@bmozaffa my understanding is the same, I think we don't need to add anything in the operators code base. Examples + docs will do, IMO.

errantepiphany:
@ricardozanini or @bmozaffa , please close this PR without merging it, as we have come to a consensus. Thank you!

PR #180, opened 2019-05-24
add product upgrade flags to CR spec

tchughesiv:
`Spec.Upgrades.Patch` defaults to true

/assign @bmozaffa 


deploy/examples/minor_upgrade.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: app.kiegroup.org/v1
+kind: KieApp
+metadata:
+  name: minor-upgrade
+spec:
+  environment: rhpam-trial
+  upgradeProductMinor: true
+  upgradeProductPatch: false
```
bmozaffa: Why show this as false in this example?

deploy/crds/kieapp.crd.yaml
```diff
@@ -605,3 +605,9 @@ spec:
                     replaceRole:
                       type: boolean
                       description: Whether to add to the current roles, or replace the current roles with the mapped ones. Replaces if set to true.
+            upgradeProductPatch:
+              type: boolean
+              description: Set false to disable automatic product patch version upgrades, it is enabled by default.
+            upgradeProductMinor:
+              type: boolean
+              description: Set true to enable automatic product minor version upgrades, it is disabled by default.
```
bmozaffa: Let's do a struct called Upgrades, and then just minor & patch in there.

deploy/examples/minor_upgrade.yaml
tchughesiv: idk, good point

deploy/crds/kieapp.crd.yaml
tchughesiv: done

deploy/examples/minor_upgrade.yaml
tchughesiv: done

PR #185, opened 2019-05-29
[RHPAM-2008] HTTP 503 error is shown in Monitoring

rhtevan:
In-sync Template changes pertaining RHPAM-2008 fix

Related PRs:
https://github.com/jboss-container-images/rhpam-7-openshift-image/pull/307
https://github.com/jboss-container-images/jboss-kie-modules/pull/251

Signed-off-by: Evan Zhang <evan.zhang@redhat.com>

### DIFF
--- a/config/common.yaml
+++ b/config/common.yaml
@@ -464,12 +464,25 @@ servers:
                 services.server.kie.org/kie-server-id: "[[.KieServerID]]"
             spec:
               serviceAccountName: "[[$.ApplicationName]]-[[$.Constants.Product]]svc"
-              terminationGracePeriodSeconds: 60
+              terminationGracePeriodSeconds: 90
               containers:
                 - name: "[[.KieName]]"
                   image: "[[$.Constants.Product]][[$.Version]]-kieserver-openshift"
                   imagePullPolicy: Always
+                  lifecycle:
+                    postStart:
+                      exec:
+                        command:
+                        - /bin/sh
+                        - /opt/eap/bin/launch/jboss-kie-kieserver-hooks.sh
+                    preStop:
+                      exec:
+                        command:
+                        - /bin/sh
+                        - /opt/eap/bin/launch/jboss-kie-kieserver-hooks.sh
                   env:
+                    - name: WORKBENCH_SERVICE_NAME
+                      value: "[[$.ApplicationName]]-[[$.Console.Name]]"
                     - name: KIE_ADMIN_USER
                       value: "[[$.AdminUser]]"
                     - name: KIE_ADMIN_PWD


tchughesiv:
@rhtevan there appears to be some issues with the yaml. Be sure you run a `make test` locally before committing. Also, have these changes been tested against an OCP cluster?

config/common.yaml
```diff
@@ -482,14 +495,6 @@ servers:
...
                     - name: DROOLS_SERVER_FILTER_CLASSES
                       value: "true"
-                    - name: KIE_SERVER_CONTROLLER_USER
-                      value: controllerUser
-                    - name: KIE_SERVER_CONTROLLER_PWD
-                      value: "[[$.ControllerPassword]]"
-                    - name: KIE_SERVER_CONTROLLER_SERVICE
-                      value: "[[$.ApplicationName]]-[[$.Console.Name]]"
-                    - name: KIE_SERVER_CONTROLLER_PROTOCOL
-                      value: ws
```
bmozaffa: Isn't OpenShift Startup Strategy optional? In other words, the user can still use the old startup strategy, no? In that case, these have to remain.

rhtevan: https://github.com/rhtevan/kie-cloud-operator/blob/f49547408accbc1a28b880d7537618f31d64e137/config/common.yaml#L491

OSSS is set. Does KieApp CR allow user to change strategy?

rhtevan: @bmozaffa Looks like the KieApp CR allows user to add arbitrary new env variables as well as overriding predefined ones. In such case, would it be cleaner if we remove those Controller related envs from the base yaml, but provide either example CR or new 'deploy/crs/.yaml' to add it back? Wdyt? For now, added those back.

bmozaffa: Hadn't you said that KIE servers still need to register with the embedded controller, for some other purposes?

rhtevan: Yes. the lifecycle hook script from KIE Server Pod, connects to BC using KIE_ADMIN_USER (Not CONTROLLER_UER) for REST/API calls.

bmozaffa: oh right, sorry if I'm repeating the same discussion again, but... this is a new dependency and we previously did not need the admin user for anything system related. Can you use controller user, or are there access issues? Again, sorry if I'm failing to remember this from last time!

rhtevan: Can we use Admin.User for now? 
Couple of reasons:
1. It is an existing user with elevated privileges configured at both Templates and Operator common.yaml right now.
2. ControllerUser/Pass has already been removed from KIE Server DC env section of all RHPAM and RHDM Templates.
3. Need a system account for connecting to BC/WB internal REST/API with sufficient privileges. At the moment Admin.User works.

bmozaffa: Yes, agree, we can for now. However, we should review this and make required improvements in 7.5. Please open a jira for yourself to this effect, targeting 7.5. Goal is to use a specific user for this purpose, most likely still called controllerUser, with only required permissions. We would then remove adminUser in SSO type of scenarios.

rhtevan: Or something like 'WBServiceUser'.

rhtevan: A follow up JIRA created (https://issues.jboss.org/projects/RHPAM/issues/RHPAM-2175)

PR #186, opened 2019-05-29
[KIECLOUD-230] - Adjust authoring-ha environment to use jboss datagrid

spolti:


config/envs/rhdm-authoring-ha.yaml
```diff
@@ -30,10 +30,18 @@ console:
                     containerPort: 8001
                     protocol: TCP
                 env:
-                  - name: APPFORMER_ELASTIC_HOST
-                    value: "[[.ApplicationName]]-[[.Constants.Product]]index"
+                  - name: OPENSHIFT_DNS_PING_SERVICE_NAME
+                    value: "[[.ApplicationName]]-[[.Console.Name]]-ping"
+                  - name: OPENSHIFT_DNS_PING_SERVICE_PORT
+                    value: "8888"
```
bmozaffa: Please look at the common.yaml and avoid overwriting anything in these files, unless you are changing the values.

config/envs/rhpam-authoring-ha.yaml
```diff
@@ -30,10 +30,18 @@ console:
                     containerPort: 8001
                     protocol: TCP
                 env:
-                  - name: APPFORMER_ELASTIC_HOST
-                    value: "[[.ApplicationName]]-[[.Constants.Product]]index"
+                  - name: OPENSHIFT_DNS_PING_SERVICE_NAME
+                    value: "[[.ApplicationName]]-[[.Console.Name]]-ping"
+                  - name: OPENSHIFT_DNS_PING_SERVICE_PORT
+                    value: "8888"
```
bmozaffa: Please look at the common.yaml and avoid overwriting anything in these files, unless you are changing the values.

config/envs/rhpam-authoring-ha.yaml
```diff
@@ -382,7 +419,7 @@ others:
                           "value": "8888"
                         }
                       ],
-                      "image": "amq-broker-72-openshift:1.3",
+                      "image": "registry.redhat.io/amq-broker-7/amq-broker-73-openshift:7.3",
```
bmozaffa: Should we put these in the constants.go file? Thoughts, @tchughesiv ?

config/jms/activemq-jms-config.yaml
```diff
@@ -65,7 +65,7 @@ servers:
                 - "[[.KieName]]-amq"
               from:
                 kind: ImageStreamTag
-                name: "amq-broker-72-openshift:1.3"
+                name: "amq-broker-73-openshift:7.3"
```
bmozaffa: Same here... @tchughesiv ?

pkg/controller/kieapp/kieapp_controller_test.go
```diff
@@ -198,60 +198,6 @@ func TestCreateRhdmImageStreams(t *testing.T) {
 	assert.Equal(t, fmt.Sprintf("registry.redhat.io/rhdm-7/rhdm%s-decisioncentral-openshift:1.0", cr.Spec.CommonConfig.Version), isTag.Tag.From.Name)
 }
 
-func TestCreateRhdmTechPreviewImageStreams(t *testing.T) {
```
tchughesiv: @spolti rather than delete these tests, should we instead modify them for datagrid images?

config/envs/rhdm-authoring-ha.yaml
```diff
@@ -383,7 +420,7 @@ others:
                           "value": "8888"
                         }
                       ],
-                      "image": "amq-broker-72-openshift:1.3",
+                      "image": "registry.redhat.io/amq-broker-7/amq-broker-73-openshift:1.3",
```
tchughesiv: hmm, i expect this will be problematic for users. this will likely require a user to setup their pam project for authenticated registry pulls... rather than allowing the DC/stateful set to first check for an imagestream in openshift project.

config/jms/activemq-jms-config.yaml
tchughesiv: might be a good idea yea

config/envs/rhdm-authoring-ha.yaml
spolti: good point, thanks.

config/envs/rhpam-authoring-ha.yaml
spolti: I think better add it do constants file.
Should I add this under TemplateConstants struct?

pkg/controller/kieapp/kieapp_controller_test.go
spolti: Agreed for the second test, as there is no thec preview image anymore, should we keep this test?

tchughesiv: i think both are valuable... one checks for latest tag, and the other checks for `1.1` versioned tag?

config/envs/rhpam-authoring-ha.yaml
```diff
@@ -290,6 +219,110 @@ others:
...
+                      value: "[[.ApplicationName]]-datagrid"
+                    - name: SERVICE_PROFILE
+                      value: "[[.ApplicationName]]-datagrid"
+                    - name: JGROUPS_PING_PROTOCOL
+                      value: openshift.DNS_PING
+                    - name: OPENSHIFT_DNS_PING_SERVICE_NAME
+                      value: "[[.ApplicationName]]-datagrid-ping"
+                    - name: INFINISPAN_CONNECTORS
+                      value: "hotrod"
+                  image: "registry.redhat.io/jboss-datagrid-7/datagrid73-openshift:1.1"
```
tchughesiv: i expect this will be problematic for users. this will likely require a user to setup their pam project for authenticated registry pulls... rather than allowing the DC/stateful set to first check for an imagestream in openshift project.

config/envs/rhpam-authoring-ha.yaml
bmozaffa: Yes, I think so, same pattern as for our own product images, I would say... and to address Tommy's other point, maybe also similarly look an image stream first as we do, and only if not found, create a local image stream pointing to source image? Thoughts, @tchughesiv ?

tchughesiv: the existing operator logic should handle it already

tchughesiv: but, yes ... needs to be changed in yaml

config/jms/activemq-jms-config.yaml
spolti: any suggestion about where should I add the constant? I mean, create a new struct or use a existing one?

config/envs/rhpam-authoring-ha.yaml
spolti: I forget to update this one.

pkg/controller/kieapp/kieapp_controller_test.go
spolti: ok, make sense.

config/jms/activemq-jms-config.yaml
tchughesiv: probably best to add a new (couple?) of strings here:
https://github.com/kiegroup/kie-cloud-operator/blob/80b536c63fa33d39d626e590961a12cc94cb91b9/pkg/apis/app/v1/kieapp_types.go#L335

something like `brokerImage` & `brokerImageTag`?
You'll have to populate this Constants section, likely in this function or another in defaults.go...
https://github.com/kiegroup/kie-cloud-operator/blob/80b536c63fa33d39d626e590961a12cc94cb91b9/pkg/controller/kieapp/defaults/defaults.go#L195

maybe create an actual constant you can use to populate w/ @
https://github.com/kiegroup/kie-cloud-operator/blob/80b536c63fa33d39d626e590961a12cc94cb91b9/pkg/controller/kieapp/constants/constants.go#L8

tchughesiv:
@bmozaffa did you want to review again before approving?

config/envs/rhpam-authoring-ha.yaml
bmozaffa: Can we move this to constants (and replace through template params)? It now appears in 8 locations, two env files and 6 unit tests. At least with the next minor update, it can be one change instead of 8 changes?

PR #191, opened 2019-05-30
Updated Operator UI to address several issues

bmozaffa:
- Previously included rename of Process Server to KIE Server for the Server ID
- [KIECLOUD-233]: Create confirmation page in the console UI
- [KIECLOUD-234]: Handling validations on field value change
- [KIECLOUD-235]: Remove unused axios dependency (Known high severity security vulnerability detected in axios <= 0.18.0)


bmozaffa:
Not merging yet, as the fix to [KIECLOUD-234] causes a regression error, in that validation can be bypassed by not navigating to the fields and deploying right away.

PR #192, opened 2019-05-31
[KIECLOUD-233] Return error to UI when deployment fails

ruromero:
Follow up on
Fix https://issues.jboss.org/browse/KIECLOUD-233

### DIFF
--- a/pkg/ui/serve.go
+++ b/pkg/ui/serve.go
@@ -4,13 +4,14 @@ package ui
 
 import (
 	"encoding/json"
+	"io/ioutil"
+
 	"github.com/RHsyseng/console-cr-form/pkg/web"
 	"github.com/ghodss/yaml"
 	"github.com/go-openapi/spec"
 	"github.com/gobuffalo/packr/v2"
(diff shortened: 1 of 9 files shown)


pkg/ui/serve.go
```diff
@@ -30,23 +31,23 @@ func Listen() {
 	}
 }
 
-func apply(cr string) {
+func apply(cr string) error {
```
bmozaffa: If you return the error, you should at most log it at debug verbosity. The caller will be able to log it, and they may have some retry logic based on the error that will mean the Error verbosity log will be confusing.

bmozaffa:
@ruromero I tried to test this by providing an invalid CR name. The error I got was 'Unexpected token . in JSON at position 346', which is not the actual deployment error, but perhaps an error while trying to parse the error response?

bmozaffa:
/hold cancel

Error reporting syntax is useful for a number of other scenarios, and even in the example case, better to know about the error than previous behavior. Merging and improving this in future releases.

pkg/ui/serve.go
bmozaffa: Will be fixed in a follow-up

PR #197, opened 2019-06-03
[KIECLOUD-231] watch/reconcile UI objects

tchughesiv:
/assign @bmozaffa 


bmozaffa:
@sutaakar this is against master, so not necessarily cherry-picked for 7.4.0, but we can apply it against 7.4.0 as well if you think appropriate. Your call!

sutaakar:
it is ok to keep it just on master, not a critical issue

PR #201, opened 2019-06-05
[KIECLOUD-97] - Synchronize OpenShiftStartupStrategy related changes …

spolti:
…between Template, APB, and Operator

Follow-up requrest, rollback the secure prefix ( it is now aligned between
templates, apb and operator)

update imagestream tag, 7.4 images uses tasg 1.0

### DIFF
--- a/config/common.yaml
+++ b/config/common.yaml
@@ -517,8 +517,6 @@ servers:
                           fieldPath: metadata.labels['services.server.kie.org/kie-server-id']
                     - name: KIE_SERVER_ROUTE_NAME
                       value: "[[.KieName]]"
-                    - name: KIE_SERVER_USE_SECURE_ROUTE_NAME
-                      value: "true"
                     - name: "[[$.Constants.MavenRepo]]_MAVEN_REPO_USERNAME"
                       value: mavenUser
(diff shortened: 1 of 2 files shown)


errantepiphany:
@spolti good catch on the image stream tag!

rhtevan:
Looks good to me.

errantepiphany:
@ruromero , @bmozaffa : Can one of you please review? Thanks!

pkg/controller/kieapp/constants/constants.go
```diff
@@ -19,7 +19,7 @@ const (
 	// ProductVersion default version
 	ProductVersion = "7.4"
 	// ImageStreamTag default tag name for the ImageStreams
-	ImageStreamTag = "1.1"
+	ImageStreamTag = "1.0"
```
bmozaffa: I think this is meant to be at 1.1 for master, and given that the change will be cherry-picked to the 7.4.0 branch, you should still get a 1.0 tag in there. @tchughesiv ?

tchughesiv: yes @bmozaffa this should remain at `1.1` for master. the cherry-pick will take care of versioning for builds.

tchughesiv: @spolti please remove this change from your PR

PR #207, opened 2019-06-13
[KIECLOUD-254] Allow default value for JNDI Name with external databases

ruromero:

Fix 
https://issues.jboss.org/projects/KIECLOUD/issues/KIECLOUD-253 with https://github.com/RHsyseng/console-cr-form/pull/82
https://issues.jboss.org/projects/KIECLOUD/issues/KIECLOUD-254 with https://github.com/RHsyseng/console-cr-form/pull/84
https://issues.jboss.org/projects/KIECLOUD/issues/KIECLOUD-255 with https://github.com/RHsyseng/console-cr-form/pull/83 and https://github.com/RHsyseng/console-cr-form/pull/85

Align with current console-ui master branch


bmozaffa:
@ruromero I don't see a fix for KIECLOUD-255. To be clear, the issue is that "KIE Server container deployment" does not have anything to do with s2i. It's about starting up a KIE server that pre-deploys a kjar. Sure, you always do that when you do s2i, but you can also do it without s2i.

ruromero:
> I don't see a fix for KIECLOUD-255. To be clear, the issue is that "KIE Server container deployment" does not have anything to do with s2i. It's about starting up a KIE server that pre-deploys a kjar. Sure, you always do that when you do s2i, but you can also do it without s2i.

@bmozaffa the Base build image has been moved before the git source configuration and the git source section is now a 0-1 object so that when the user wants to use s2i the git uir and reference can become mandatory.
This change together with renaming the section should be enough. Shouldn't it?

deploy/crds/kieapp.crd.yaml
```diff
@@ -363,9 +362,6 @@ spec:
                               nonXA:
                                 type: string
                                 description: Sets the datasources type. It can be XA or NONXA. For non XA set it to true. Default value is false.
-                              jndiName:
-                                type: string
-                                description: Database JNDI name used by application to resolve the datasource, e.g. java:/jboss/datasources/ExampleDS
```
bmozaffa: We can't make changes like this to the CRD, specially not properties.

deploy/crds/kieapp.1.0.crd.yaml
```diff
@@ -354,9 +353,6 @@ spec:
                               nonXA:
                                 type: string
                                 description: Sets the datasources type. It can be XA or NONXA. For non XA set it to true. Default value is false.
-                              jndiName:
-                                type: string
-                                description: Database JNDI name used by application to resolve the datasource, e.g. java:/jboss/datasources/ExampleDS
```
bmozaffa: This is the already released version of the CRD, and will never change.

pkg/apis/app/v1/kieapp_types.go
```diff
@@ -314,7 +314,6 @@ type ExternalDatabaseObject struct {
 	Port                       string `json:"port,omitempty"`
 	JdbcURL                    string `json:"jdbcURL,omitempty"`
 	NonXA                      string `json:"nonXA,omitempty"`
-	JndiName                   string `json:"jndiName,omitempty"`
```
bmozaffa: See above

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -502,6 +502,7 @@ func getDatabaseConfig(environment v1.EnvironmentType, database *v1.DatabaseObje
 	if database.Type == v1.DatabaseExternal && database.ExternalConfig == nil {
 		return nil, fmt.Errorf("external database configuration is mandatory for external database type")
 	}
+
```
bmozaffa: Let's not get this file in the change list for a whitespace change

ruromero:
> See comments, CRD changes are very complicated and I didn't realize what I was implying when I said that about the console... maybe just do the change there? Thoughts, @tchughesiv ?

It will fail for external DB because in the spec the jndiName is required

deploy/crds/kieapp.crd.yaml
ruromero: Maybe just remove it from the list of required properties and keep it?

bmozaffa:
Closed in favor of [PR 209](https://github.com/kiegroup/kie-cloud-operator/pull/209)

PR #212, opened 2019-06-24
[KIECLOUD-149] Fix rhdm/rhpam trial environments

ruromero:
Fix https://issues.jboss.org/browse/KIECLOUD-149

* Disable JGROUPS for DecisionManager and BusinessCentral in trial environments

### DIFF
--- a/config/envs/rhdm-trial.yaml
+++ b/config/envs/rhdm-trial.yaml
@@ -7,6 +7,11 @@ console:
           spec:
             containers:
               - name: "[[.ApplicationName]]-[[.Console.Name]]"
+                env:
+                  - name: KIE_SERVER_HOST
+                  - name: JGROUPS_PING_PROTOCOL
+                  - name: OPENSHIFT_DNS_PING_SERVICE_NAME
+                  - name: OPENSHIFT_DNS_PING_SERVICE_PORT
                 ports:
                   - name: git-ssh
                     containerPort: 8001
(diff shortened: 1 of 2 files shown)


config/envs/rhdm-trial.yaml
```diff
@@ -7,6 +7,11 @@ console:
           spec:
             containers:
               - name: "[[.ApplicationName]]-[[.Console.Name]]"
+                env:
+                  - name: KIE_SERVER_HOST
```
sutaakar: What is the purpose of specifying KIE_SERVER_HOST into console deployment config?

config/envs/rhpam-trial.yaml
```diff
@@ -7,6 +7,11 @@ console:
           spec:
             containers:
               - name: "[[.ApplicationName]]-[[.Console.Name]]"
+                env:
+                  - name: KIE_SERVER_HOST
```
sutaakar: Same as above.

config/envs/rhdm-trial.yaml
ruromero: Setting it without value will override the value from the common.yaml file

sutaakar: Ok, thanks

sutaakar: But why is it specified for Workbench? AFAIK Workbench has nothing to do with KIE_SERVER_HOST.

ruromero: Because it's wrong and shouldn't be there :wink:

PR #217, opened 2019-07-15
[RHPAM-1387] - Allow set Smart router location to external URL in RHPAM

spolti:


config/common.yaml
```diff
@@ -349,6 +350,12 @@ smartRouter:
                     value: "kie-server-router"
                   - name: KIE_SERVER_ROUTER_NAME
                     value: "KIE Server Router"
+                  - name: KIE_SERVER_ROUTER_PROTOCOL
+                    value: "https"
```
tchughesiv: This is set to http elsewhere in the Kie DC, should they be set the same?

PR #225, opened 2019-07-22
[KIECLOUD-277] - JMS Parameters available for user to configure but n…

spolti:
…ot applied on kieserver template.

While adding the two missing parameters to the template was identified if, when a default value of
the JMS boolean parameter is true, if this parameter is empty or false, the default will be true.
To fix that, boolean pointers was used, in that way when the param is empty or false, it will be
correctly set.


tchughesiv:
/lgtm
/approve

@spolti i assume a cherry-picked PR against master is on its way as well?

spolti:
@tchughesiv yes.

tchughesiv:
@spolti could you explain this sentence a little more?
_when a default value of the JMS boolean parameter is true, if this parameter is empty or false, the default will be true._

if the parameter is empty or false, the default will be true... in the go template? java product code?

spolti:
@tchughesiv, sure, for example,it happens in the go template.
 For example, if you define one of the JMS parameters which is defaulted to true ( KieServerJmsExecutor and KieServerJmsAuditTransacted), when we merge the two structs with mergo, these 2 values, even if defined as false, will be redefined to true.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -545,17 +549,19 @@ func getJmsConfig(environment v1.EnvironmentType, jms *v1.KieAppJmsObject) (*v1.
 		KieServerJmsQueueSignal:        "queue/KIE.SERVER.SIGNAL",
 		KieServerJmsEnableAudit:        false,
 		KieServerJmsQueueAudit:         "queue/KIE.SERVER.AUDIT",
-		KieServerJmsAuditTransacted:    true,
+		KieServerJmsAuditTransacted:    t,
 		KieServerJmsUsername:           "user" + string(shared.GeneratePassword(4)),
 		KieServerJmsPassword:           string(shared.GeneratePassword(8)),
 	}
 
+	mergo.Merge(defaultJms, jms, mergo.WithOverride)
```
tchughesiv: @spolti i wonder if the addition of this line is what's causing the template problem?

mergo is already executed a few lines later...
https://github.com/kiegroup/kie-cloud-operator/blob/4a610088e81903d79fb1151a52a7ff910878df46/pkg/controller/kieapp/defaults/defaults.go#L576

spolti: Yes, I did this to firstly, merge the jms values on to defaultJms values overriding all values set.
the second is used only the apply the Queues parameter and, if I return the defaultJms instead the jms struct, the operator didn't work, the AMQ pod never gets up. not sure why.

bmozaffa: @spolti to figure this out, would you please create a new branch, and in there, create a unit test that reproduces the original problem with the true/false default? Let us look at that to understand the problem, and if this is the best/right solution.

spolti:
@tchughesiv the PR is updated with your suggestions, thanks!

pkg/controller/kieapp/defaults/defaults.go
spolti: @bmozaffa a small reproducer can be found with the changes on this branch: https://github.com/spolti/kie-cloud-operator/tree/temp

PR #234, opened 2019-07-29
[KIECLOUD-281] Change operator to use versioned config dirs

tchughesiv:
@bmozaffa thoughts on this approach?


bmozaffa:
> @bmozaffa thoughts on this approach?

I was thinking about this and came up with the exact same approach, so yes, it looks great!

PR #245, opened 2019-08-07
[KIECLOUD-293] add api versioning capability and product upgrade/version checks

tchughesiv:
- incremented to v2 api
 - switched to `api` import alias for latest api version
 - switched to versioned schema validation in crd
 - versioned cr examples directories
 - moved product `version` field to base of spec; allows OLM to display
 - added `phase` field to status; allows OLM to display
 - added boolean switches for upgrade to OLM csv spec descriptors & form.json; allows users to opt-in from UIs

/assign @bmozaffa @ruromero 

Screen Shot 2019-08-12 at 7 04 04 PM
Screen Shot 2019-08-07 at 4 56 38 PM
Screen Shot 2019-08-09 at 4 49 37 PM


deploy/catalog_resources/community/1.2.0/kiecloud-operator.1.2.0.clusterserviceversion.yaml
```diff
@@ -99,15 +99,26 @@ spec:
...
           - description: Product version installed.
             displayName: Version
-            path: commonConfig.version
+            path: version
             x-descriptors:
               - "urn:alm:descriptor:com.tectonic.ui:label"
+          - description: Specifies whether or not to execute product upgrades. When true, assumes both patch and minor upgrades unless set otherwise in CR.
+            displayName: Allow Upgrades
+            value: false
+            path: upgrades.minor
```
bmozaffa: Why expose this and not patch? In other words, what if they only watch patch upgrades?

tchughesiv: I went back and forth on this, but ultimately there were multiple reasons i went with only minor for UI exposure.
 - While minor is a normal boolean, patch is setup as a pointer. This is so that we can default to true for patch whenever minor is true, but still allows a user to explicitly set patch to false or only opt in for patch upgrades. This is problematic when enabling patch in the UI because if a user clicks the patch upgrade buttom/checkbox and then turns it off, patch is now set to false instead of nil. The operator will assume the user has chosen NOT to install patch upgrades.
 - Simplicity, one button is better than two if it enables upgrades as expected. remember, enabling minor also assumes patch is true unless patch is explicitly set to false.

tchughesiv: Alternative is, we change patch to standard boolean and move away from defaulting to true behavior when minor true. This might make things clearer, allow for both flags in UI, and reduce chance of user confusion around upgrades.

current code:
https://github.com/kiegroup/kie-cloud-operator/blob/d5a7045f073af44d8b56856c6871c05ee5b2d8fa/pkg/controller/kieapp/defaults/upgrade.go#L13-L28

bmozaffa: Does it make sense to ever say you want to upgrade to a minor version, but not to a micro version? I think not, but maybe check with Alessandro as well?

If not:

- How about we have one flag for "allow automatic upgrades" and another one for "upgrade minor versions", where the first assumes patch and both are required for minor to happen

Is yes:

- Have separate booleans for micro and minor, and no implicit behavior... require both to be set to take effect.

lazarotti: I dont see any reason why someone could be interest to auto-upgrade to minors, but not to micros (path updates). The opposite is true, someone could set micro upgrades but not minors.
Not sure if understood completely your first option Babak, but I would suggest something similar, but being more clear into display name the differences:
1)"Allow automatic upgrades for patches and minor releases" 
2)"Allow automatic upgrades for patches only"

... or this will bring again the issue if user check/uncheck the second option (it will break the first logic that applies for both)

bmozaffa: Cool, so going with "not"... 

We could use the long sentences you mention in labels, but the main syntax is yaml. We need concise names that convey the meeting. Your two options are the same as mine, but expressed in reverse order. If I were to rephrase mine, I suggested:

1- Enable upgrades (includes micro by default)
2- Include minor versions in upgrades

tchughesiv: Following @lazarotti's suggestion, changing spec and UI fields accordingly. I should be able to make this work.
```
upgrades:
  minorMicro:
  micro:
```

tchughesiv: @lazarotti ... other option could be... as @bmozaffa stated:
```
upgrades:
  enabled:
  minor:
```
flow would be... when enabled, micro updates are turned on... user would then have to opt in further for minor upgrades. UI options might look more like:
1. "Enable Upgrades"
2. "Allow automatic upgrades minor releases"

option 2 doesn't work unless option 1 is true

tchughesiv:
@bmozaffa changes completed including UI improvements for both marketplace and console form. ready for review.

PR #249, opened 2019-08-19
[KIECLOUD-289] custom image/tag support for objects

tchughesiv:
/assign @bmozaffa 

### DIFF
--- a/deploy/examples/v2/custom_images_registry.yaml
+++ b/deploy/examples/v2/custom_images_registry.yaml
@@ -1,0 +1,21 @@
+apiVersion: app.kiegroup.org/v2
+kind: KieApp
+metadata:
+  name: image-registry
+spec:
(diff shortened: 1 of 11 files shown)


deploy/examples/v2/custom_images_registry.yaml
```diff
@@ -0,0 +1,21 @@
...
+    registry: registry.example.com:5000
+  objects:
+    console:
+      image: console-image
+      imageTag: "1.2"
+    servers:
+      - image: server-image
+        imageTag: "1.2"
+      - image: server-image2
+        imageTag: "1.1"
```
bmozaffa: What prompted this change? I would not worry about trying to support two different versions of KIE server in the same deployment, if that's the motivation. I see that it adds the flexibility to use custom image names; is that the reason?

PR #265, opened 2019-09-05
[RHPAM-2354] OutOfMemoryError on Datagrid cause issues on project import indexing

errantepiphany:
[RHPAM-2354] OutOfMemoryError on Datagrid cause issues on project import indexing
https://issues.jboss.org/browse/RHPAM-2354

Signed-off-by: David Ward <dward@redhat.com>

### DIFF
--- a/pkg/controller/kieapp/constants/constants.go
+++ b/pkg/controller/kieapp/constants/constants.go
@@ -76,7 +76,7 @@ var VersionConstants = map[string]*api.VersionConfigs{
 		BrokerImage:      "amq-broker-73-openshift",
 		BrokerImageTag:   "7.3",
 		DatagridImage:    "datagrid73-openshift",
-		DatagridImageTag: "1.1",
+		DatagridImageTag: "1.2",
 	},
 	LastMicroVersion: {
 		APIVersion:       v1.SchemeGroupVersion.Version,
--- a/config/7.5.0/envs/rhdm-authoring-ha.yaml
+++ b/config/7.5.0/envs/rhdm-authoring-ha.yaml
@@ -277,10 +277,11 @@ others:
                     timeoutSeconds: 10
                   resources:
                     limits:
-                      memory: "512Mi"
+                      cpu: "1000m"
+                      memory: "2Gi"
                     requests:
-                      cpu: "0.5"
-                      memory: "512Mi"
+                      cpu: "1000m"
+                      memory: "2Gi"
                   volumeMounts:
                     - mountPath: /opt/datagrid/standalone/data
                       name: srv-data
--- a/config/7.5.0/envs/rhpam-authoring-ha.yaml
+++ b/config/7.5.0/envs/rhpam-authoring-ha.yaml
@@ -277,10 +277,11 @@ others:
                     timeoutSeconds: 10
                   resources:
                     limits:
-                      memory: "512Mi"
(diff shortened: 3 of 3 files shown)


errantepiphany:
Please do **not** merge as this is a work-in-progress. I know that the cpu limit is invalid, and that the memory limit is too high. They need to get fixed in https://github.com/jboss-container-images/rhdm-7-openshift-image/pull/284 and https://github.com/jboss-container-images/rhpam-7-openshift-image/pull/357 too, but the correct values are still under discussion.

I also need to verify that my DatagridImageTag bump from 1.1 to 1.2 actually belongs in all of CurrentVersion, LastMicroVersion, and LastMinorVersion. @tchughesiv , do you know?

pkg/controller/kieapp/constants/constants.go
```diff
@@ -76,23 +76,23 @@ var VersionConstants = map[string]*api.VersionConfigs{
...
+		DatagridImageTag: "1.2",
 	},
 	LastMicroVersion: {
 		APIVersion:       v1.SchemeGroupVersion.Version,
 		ImageTag:         "1.1",
 		BrokerImage:      "amq-broker-73-openshift",
 		BrokerImageTag:   "7.3",
 		DatagridImage:    "datagrid73-openshift",
-		DatagridImageTag: "1.1",
+		DatagridImageTag: "1.2",
```
tchughesiv: looks like you're targeting only version 7.5.0 for this, but with this imagetag change... you're affecting 7.4.1 as well. You should likely only modify the "CurrentVersion" mapping, as that reflects 7.5.0.
https://github.com/kiegroup/kie-cloud-operator/blob/32ea6118a75f11c9d48362578127608f10e8efa6/pkg/controller/kieapp/constants/constants.go#L9-L16

pkg/controller/kieapp/constants/constants.go
```diff
@@ -76,23 +76,23 @@ var VersionConstants = map[string]*api.VersionConfigs{
...
 	},
 	LastMinorVersion: {
 		APIVersion:       v1.SchemeGroupVersion.Version,
 		ImageTag:         "1.0",
 		BrokerImage:      "amq-broker-73-openshift",
 		BrokerImageTag:   "7.3",
 		DatagridImage:    "datagrid73-openshift",
-		DatagridImageTag: "1.1",
+		DatagridImageTag: "1.2",
```
tchughesiv: same, but for 7.4.0 ... as a reference, here's where we define which version is which...
https://github.com/kiegroup/kie-cloud-operator/blob/32ea6118a75f11c9d48362578127608f10e8efa6/pkg/controller/kieapp/constants/constants.go#L9-L16

pkg/controller/kieapp/constants/constants.go
errantepiphany: Got it; thanks. I'll force-push the fix. Much appreciated. :)

pkg/controller/kieapp/constants/constants.go
errantepiphany: This mapping helps it make sense. Thanks again.

errantepiphany:
Modifications made to match what we're going with in the templates. Ready for review, approval, and merge now. @tchughesiv , can you please take a look? Thanks!

PR #267, opened 2019-09-06
[RHDM-1057_RHPAM-2279] - Review and update Authoring HA template for …

spolti:
…latest changes


config/7.5.0/envs/rhdm-authoring-ha.yaml
```diff
@@ -46,165 +46,6 @@ console:
 # ES/AMQ BEGIN
 others:
   - roles:
```
tchughesiv: any reason this roles array is still here at all? looks like you deleted everything underneath it?

config/7.5.0/envs/rhpam-authoring-ha.yaml
```diff
@@ -46,165 +46,6 @@ console:
 # ES/AMQ BEGIN
 others:
   - roles:
```
tchughesiv: same...

config/7.5.0/envs/rhdm-authoring-ha.yaml
```diff
@@ -474,7 +212,7 @@ others:
                         fieldRef:
                           apiVersion: v1
                           fieldPath: metadata.namespace
-                  image: "registry.redhat.io/amq-broker-7/[[$.Constants.BrokerImage]]:[[$.Constants.BrokerImageTag]]"
+                  image: "registry.redhat.io/amq7/[[$.Constants.BrokerImage]]:[[$.Constants.BrokerImageTag]]"
```
errantepiphany: @spolti : Don't we now have to add variable here for the first part of the image name?

"amq-broker-7" is fine for LastMinorVersion and LastMicroVersion here: https://github.com/kiegroup/kie-cloud-operator/blob/fab4d6a8f381bf3e2a21ea8ebd744b6aa3d45144/pkg/controller/kieapp/constants/constants.go#L73-L96 , however it is "amq7" in CurrentVersion. So, I think we'd have to come up with another variable, yes? Perhaps something like [[$.Constants.BrokerImageProduct]]?

If so, the line would be:
image: "registry.redhat.io/[[$.Constants.BrokerProduct]]/[[$.Constants.BrokerImage]]:[[$.Constants.BrokerImageTag]]"

@tchughesiv , wdyt?

tchughesiv: @errantepiphany its a thought but likely unnecessary as the config templates themselves are versioned, so 7.4.1 can have a different context setting in the templates themselves... e.g. -
https://github.com/kiegroup/kie-cloud-operator/blob/fab4d6a8f381bf3e2a21ea8ebd744b6aa3d45144/config/7.4.1/envs/rhdm-authoring-ha.yaml#L490

config/7.5.0/envs/rhdm-authoring-ha.yaml
spolti: If I remove that this issue happens:

`--- FAIL: TestRhdmAuthoringHAEnvironment (0.05s)
    defaults_test.go:212: 
                Error Trace:    defaults_test.go:212
                Error:          Expected nil, but got: &errors.errorString{s:"error unmarshaling JSON: json: cannot unmarshal object into Go struct field Environment.others of type []v2.CustomObject"}
                Test:           TestRhdmAuthoringHAEnvironment
                Messages:       Error getting prod environment
panic: runtime error: index out of range [recovered]
        panic: runtime error: index out of range
`

spolti: maybe a indentation error?

config/7.5.0/envs/rhdm-authoring-ha.yaml
errantepiphany: @tchughesiv Ah, yeah, you're right. /facepalm. I totally missed that the config directory of @spolti 's change was itself "7.5.0". Nevermind!

errantepiphany: (Resolving this conversation as I stand corrected.)

config/7.5.0/envs/rhdm-authoring-ha.yaml
tchughesiv: Leave the dash only

PR #285, opened 2019-10-16
move away from role wildcard rules

tchughesiv:
The reason for this is to further restrict the permissions we're allotting to our operator/product. This should allow us to offer dedicated users a way to deploy our operator/product at some point.

But also, this allows us to know what we really need. My initial testing has been successful.

### DIFF
--- a/config/7.6.0/common.yaml
+++ b/config/7.6.0/common.yaml
@@ -837,14 +837,31 @@ others:
         rules:
           - apiGroups:
               - ""
-              - app.kiegroup.org
-              - apps.openshift.io
-              - image.openshift.io
-              - route.openshift.io
             resources:
-              - "*"
+              - configmaps
+              - serviceaccounts
             verbs:
-              - "*"
+              - create
+              - delete
+              - deletecollection
+              - get
+              - list
+              - patch
+              - update
+              - watch
+          - apiGroups:
+              - apps.openshift.io
+            resources:
+              - deploymentconfigs
+            verbs:
+              - create
+              - delete
+              - deletecollection
+              - get
+              - list
+              - patch
+              - update
+              - watch
 
     serviceAccounts:
       - metadata:
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -5,17 +5,117 @@ metadata:
 rules:
 - apiGroups:
   - ""
+  resources:
+  - configmaps
+  - pods
+  - services
+  - serviceaccounts
+  - persistentvolumeclaims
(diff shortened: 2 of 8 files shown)


deploy/role.yaml
```diff
@@ -28,7 +126,14 @@ rules:
...
   verbs:
-  - '*'
+  - create
+  - delete
+  - deletecollection
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
bmozaffa: Can we limit these from CSVs?

tchughesiv: likely yes...

tchughesiv: is this better @bmozaffa ?

bmozaffa: Great, thanks!

tchughesiv:
/hold
@errantepiphany @rhtevan any issues with these changes? specifically the ones in the environment template common.yaml files.

The reason for this is to further restrict the permissions we're allotting to our operator/product. This should allow us to offer dedicated users a way to deploy our operator/product at some point.

But also, this allows us to know what we really need. Including @rhtevan @bmozaffa in case further discussion is necessary.

config/7.6.0/common.yaml
```diff
@@ -837,14 +837,75 @@ others:
...
           - apiGroups:
               - ""
-              - app.kiegroup.org
+            resources:
+              - configmaps
+              - pods
+              - services
+              - serviceaccounts
+              - persistentvolumeclaims
+              - secrets
```
tchughesiv: my main question here is for @errantepiphany @rhtevan  ... does PAM need all of this access? I've put this here instead of the wildcard access we are currently declaring, but it still seems excessive. 

Specifically, what does PAM need writes to beyond view/edit rights for ConfigMaps?  Are there any other openshift/k8s objects that PAM interacts with at runtime?

rhtevan: AFAIK, PAM needs to be able to manipulate ConfigMap, watch and trigger rollout for DeploymentConfig, but not interacts with Secrets. PAM image scripts interact with KubeAPI server for querying.

tchughesiv: so i can remove the other resource types from this Role? Just leave ConfigMaps and DeploymentConfigs?

@rhtevan @errantepiphany  ?

rhtevan: I would say at least keep 'serviceaccounts'. @errantepiphany  anything else needed from script perspective?

PR #287, opened 2019-10-18
[RHPAM-2401] - Change strategy to type Rolling for other BA components

spolti:

### DIFF
--- a/config/7.6.0/common.yaml
+++ b/config/7.6.0/common.yaml
@@ -304,7 +304,10 @@ smartRouter:
           service: "[[.ApplicationName]]-smartrouter"
       spec:
         strategy:
-          type: Recreate
+          rollingParams:
+            maxSurge: 100%
+            maxUnavailable: 0
+          type: Rolling
         triggers:
           - type: ImageChange
             imageChangeParams:
--- a/config/7.6.0/dbs/postgresql.yaml
+++ b/config/7.6.0/dbs/postgresql.yaml
@@ -63,7 +63,10 @@ servers:
             service: "[[.KieName]]-postgresql"
         spec:
           strategy:
-            type: Recreate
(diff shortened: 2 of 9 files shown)


config/7.5.1/common.yaml
```diff
@@ -7,8 +7,6 @@ console:
           application: "[[.ApplicationName]]"
           service: "[[.ApplicationName]]-[[.Console.Name]]"
       spec:
-        strategy:
-          type: Recreate
```
tchughesiv: @spolti instead of removing Recreate from common and adding everywhere else... maybe leave common as-is and just change the strategy in the couple specific envs that need to be rolling.

spolti: I tried that, I got it as recreate strategy when I ran the tests.

errantepiphany: @tchughesiv What would you suggest give what @spolti experienced above? ^^

tchughesiv:
@spolti please rebase and move the changes to 7.6.0 config files now in master branch

pkg/apis/app/v2/zz_generated.deepcopy.go
```diff
@@ -1,6 +1,6 @@
 // +build !ignore_autogenerated
 
-// Code generated by deepcopy-gen. DO NOT EDIT.
+// Code generated by operator-sdk. DO NOT EDIT.
```
tchughesiv: Please be sure using sdk 0.7.0 and fix commit

config/7.5.1/common.yaml
tchughesiv: @spolti @errantepiphany for example...
https://github.com/spolti/kie-cloud-operator/pull/1

PR #310, opened 2019-11-27
[KIECLOUD-343] allow product to get routes

tchughesiv:
/assign @sutaakar @bmozaffa 

### DIFF
--- a/config/7.5.0/common.yaml
+++ b/config/7.5.0/common.yaml
@@ -792,6 +792,19 @@ others:
               - patch
               - update
               - watch
+          - apiGroups:
+              - route.openshift.io
+            resources:
+              - routes
+            verbs:
+              - create
+              - delete
+              - deletecollection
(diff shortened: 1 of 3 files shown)


tchughesiv:
@sutaakar @errantepiphany while this PR adds back route access for the product... i've noticed we also removed image.openshift.io group access in that prior change...
https://github.com/kiegroup/kie-cloud-operator/pull/285/files#diff-a5c0693ab31d2c16b6fa64a4d1df72acL772

is there any reason you can think of that this access should be added back... and if so, which resource is needed? imagestreams? imagestreamtags?

bmozaffa:
/approve cancel
/lgtm
@tchughesiv Removing approve label so you can merge when ready, after discussion has concluded.

sutaakar:
I am not aware about any needs to have access rights to images.

PR #342, opened 2020-01-23
[RHPAM-2175] Use only adminUser

ruromero:

Fix https://issues.redhat.com/browse/RHPAM-2175

### DIFF
--- a/deploy/crds/kieapp.crd.yaml
+++ b/deploy/crds/kieapp.crd.yaml
@@ -770,15 +770,6 @@ spec:
                     amqClusterPassword:
                       type: string
                       description: The password to use for amq cluster user.
-                    controllerPassword:
-                      type: string
-                      description: The password to use for the controllerUser.
-                    serverPassword:
-                      type: string
-                      description: The password to use for the executionUser.
-                    mavenPassword:
-                      type: string
-                      description: The password to use for the mavenUser.
                 objects:
                   type: object
                   description: Configuration of the RHPAM components
--- a/pkg/apis/app/v2/kieapp_types.go
+++ b/pkg/apis/app/v2/kieapp_types.go
@@ -450,9 +450,12 @@ type CommonConfig struct {
 	DBPassword         string `json:"dbPassword,omitempty"`
 	AMQPassword        string `json:"amqPassword,omitempty"`
 	AMQClusterPassword string `json:"amqClusterPassword,omitempty"`
+	// Deprecated - Remove after 7.6.1 is also removed
 	ControllerPassword string `json:"controllerPassword,omitempty"`
-	ServerPassword     string `json:"serverPassword,omitempty"`
-	MavenPassword      string `json:"mavenPassword,omitempty"`
+	// Deprecated - Remove after 7.6.1 is also removed
+	ServerPassword string `json:"serverPassword,omitempty"`
+	// Deprecated - Remove after 7.6.1 is also removed
+	MavenPassword string `json:"mavenPassword,omitempty"`
 }
 
 // VersionConfigs ...
--- a/pkg/apis/app/v2/kieapp_types.go
+++ b/pkg/apis/app/v2/kieapp_types.go
@@ -450,11 +450,11 @@ type CommonConfig struct {
 	DBPassword         string `json:"dbPassword,omitempty"`
 	AMQPassword        string `json:"amqPassword,omitempty"`
 	AMQClusterPassword string `json:"amqClusterPassword,omitempty"`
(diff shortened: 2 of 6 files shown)


pkg/apis/app/v2/kieapp_types.go
```diff
@@ -450,9 +450,12 @@ type CommonConfig struct {
 	DBPassword         string `json:"dbPassword,omitempty"`
 	AMQPassword        string `json:"amqPassword,omitempty"`
 	AMQClusterPassword string `json:"amqClusterPassword,omitempty"`
+	// Deprecated - Remove after 7.6.1 is also removed
```
spolti: we will not make 7.6.1, can it be removed now?

tchughesiv: I'd imagine this needs to stay until 7.6.0 rolls off. We still support both 7.6 and 7.5.1 installs with the 7.7 operator.

deploy/crds/kieapp.crd.yaml
```diff
@@ -770,15 +770,6 @@ spec:
...
                       description: The password to use for amq cluster user.
-                    controllerPassword:
-                      type: string
-                      description: The password to use for the controllerUser.
-                    serverPassword:
-                      type: string
-                      description: The password to use for the executionUser.
-                    mavenPassword:
-                      type: string
-                      description: The password to use for the mavenUser.
```
tchughesiv: @ruromero @bmozaffa 
More a conversation starter than anything, should we version our api for adds/removes or only "changes" to existing components?

bmozaffa: I think we should decide case by case, with a major driver of our decision being the difficulty and complication of version increments, which I hope will get better. In fact the level of difficulty should even affect our decision to make CRD changes.

As long as CRD versioning remains as difficult as it has been, I would advocate to only version when there is a change, or required field is removed, or similar. New critical and required fields may be an exception, if we want to make sure they are exposed in the CRD.

ruromero: In this case the remove fields are not required and most of the times generated. I don't consider this a breaking change but that's something to always have in mind and agree on beforehand.
Thanks

pkg/apis/app/v2/kieapp_types.go
ruromero: @spolti I don't want to remove it until version 7.7 is the last one supported by the operator to allow users still deploy v1 of the CRD.

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -450,9 +450,12 @@ type CommonConfig struct {
 	DBPassword         string `json:"dbPassword,omitempty"`
 	AMQPassword        string `json:"amqPassword,omitempty"`
 	AMQClusterPassword string `json:"amqClusterPassword,omitempty"`
+	// Deprecated - Remove when 7.7.0 is the oldest version supported
```
ruromero: @spolti I think this comment is clearer

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -636,10 +636,7 @@ func setPasswords(cr *api.KieApp, isTrialEnv bool) {
 		&cr.Spec.CommonConfig.AdminPassword,
 		&cr.Spec.CommonConfig.DBPassword,
 		&cr.Spec.CommonConfig.AMQPassword,
-		&cr.Spec.CommonConfig.AMQClusterPassword,
-		&cr.Spec.CommonConfig.ControllerPassword,
-		&cr.Spec.CommonConfig.MavenPassword,
-		&cr.Spec.CommonConfig.ServerPassword}
```
tchughesiv: do we need to add some logic here that still adds these passwords if the version is less than 7.7.0? I assume we'll still need these passwords for older product versions?

maybe something like
```shell
	if getMinorImageVersion(cr.Spec.Version) < "77" {
               passwords = append(passwords, &cr.Spec.CommonConfig.ControllerPassword) ...
....
        }
```
???

ruromero: The templates for 7.5.1 and 7.6.0 have the variables set so I think we can just delete them once the oldest templates are 7.7.0

tchughesiv: i think that makes sense ya... so, for this change then, would you add them back to this slice or?

ruromero: Of course, I totally forgot.

PR #345, opened 2020-01-27
add CI that catches prior version config changes

tchughesiv:


tchughesiv:
/assign @bmozaffa 
view failed unit test for result of this new check

tchughesiv:
should this be split out as a separate test or do we keep as part of unit?

PR #349, opened 2020-01-29
[RHPAM-2658] Define value as byte

ruromero:

Fix https://issues.redhat.com/browse/RHPAM-2658

I have reproduced the issue and validated the fix on the following environment

```
Server Version: 4.4.0-0.nightly-2020-01-23-054055
Kubernetes Version: v1.17.1
```

```
oc create -f deploy/catalog_resources/redhat/1.4.0/businessautomation-operator.1.4.0.clusterserviceversion.yaml                                                                                                  
The ClusterServiceVersion "businessautomation-operator.1.4.0" is invalid: spec.customresourcedefinitions.owned.specDescriptors.value: Invalid value: "": spec.customresourcedefinitions.owned.specDescriptors.value in body must be of type byte: ""
```

### DIFF
--- a/deploy/catalog_resources/community/1.2.0/kiecloud-operator.1.2.0.clusterserviceversion.yaml
+++ b/deploy/catalog_resources/community/1.2.0/kiecloud-operator.1.2.0.clusterserviceversion.yaml
@@ -94,13 +94,15 @@ spec:
         specDescriptors:
           - description: Set true to enable automatic micro version product upgrades, it is disabled by default.
             displayName: Enable Upgrades
-            value: false
+            value:
+              - false
             path: upgrades.enabled
             x-descriptors:
               - "urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
           - description: Set true to enable automatic minor product version upgrades, it is disabled by default. Requires spec.upgrades.enabled to be true.
             displayName: Include minor version upgrades
-            value: false
+            value:
+              - false
             path: upgrades.minor
             x-descriptors:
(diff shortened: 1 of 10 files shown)


tchughesiv:
@ruromero please try to persist this change in `tools/csv-gen`
https://github.com/kiegroup/kie-cloud-operator/blob/8d246ac54f4604d594b5325015257e56bb283b19/tools/csv-gen/csv-gen.go#L224

ruromero:
@tchughesiv I preferred to do it when the value is generated from a string to make it more reusable

madorn:
@ruromero @tchughesiv  - hi folks, just a note that just discovered boolean switch will automatically set this false by default (unless otherwise set to "true" in the CR). It shouldn't be necessary to explicitly set false.

tchughesiv:
thanks @madorn ... thoughts no longer setting to false @ruromero ?

ruromero:
We can remove the flags but keep the change in the generator. Does that sound right?

bmozaffa:
Right, though I don't see anything wrong with providing default values either... if anything, it's maybe more explicit and obvious to someone reviewing it.

madorn:
just discovered this change is not backwards compatible with 4.2:

* Completely omitting specDescriptors.value in 4.2 will cause booleanUI button to show true.

* Using specDescriptors.value = [false] in 4.2 will also show true.

* Using specDescriptors.value = "false" in 4.2 will show false.

Consulting with dev team. Please stand by.

PR #351, opened 2020-01-30
Revert "[RHPAM-2658] Define value as byte"

tchughesiv:
Reverts kiegroup/kie-cloud-operator#349 given @madorn's findings... can address again once a solution is in place for ocp 4.4
https://github.com/kiegroup/kie-cloud-operator/pull/349#issuecomment-580415423

### DIFF
--- a/deploy/catalog_resources/community/1.2.0/kiecloud-operator.1.2.0.clusterserviceversion.yaml
+++ b/deploy/catalog_resources/community/1.2.0/kiecloud-operator.1.2.0.clusterserviceversion.yaml
@@ -94,11 +94,13 @@ spec:
         specDescriptors:
           - description: Set true to enable automatic micro version product upgrades, it is disabled by default.
             displayName: Enable Upgrades
+            value: false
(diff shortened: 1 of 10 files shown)


tchughesiv:
@ruromero @bmozaffa

madorn:
@tchughesiv 

Here's the final word on 4.4 compatibility:

As of OpenShift 4.4,specDescriptors[].value in CSVs will need to be omitted.

The two options to resolve are:

* Move those values into defaults in your CRD: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting. The caveat with option is that CRD defaulting requires `apiVersion: apiextensions.k8s.io/v1`
for `kind: CustomResourceDefinition`. This API endpoint is not available in OpenShift 3.11 but is present in OpenShift 4.0 and greater.

* Move the desired values to the Custom Resource reflected in the `alm-examples` section of your ClusterServiceVersion.

If I can assist you with any of this, please let me know.

PR #354, opened 2020-02-04
[BAPL-1507] Add to OpenShift images property for JVM max memory ratio configuration

errantepiphany:
[BAPL-1507] Add to OpenShift images property for JVM max memory ratio configuration
https://issues.redhat.com/browse/BAPL-1507

Signed-off-by: David Ward <dward@redhat.com>


pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -623,6 +619,83 @@ func testAMQEnvs(t *testing.T, kieserverEnvs []corev1.EnvVar, amqEnvs []corev1.E
...
+		case "GC_TIME_RATIO":
+			assert.Equal(t, "4", env.Value)
+
+		case "GC_ADAPTIVE_SIZE_POLICY_WEIGHT":
+			assert.Equal(t, "90", env.Value)
+
+		case "GC_MAX_METASPACE_SIZE":
+			assert.Equal(t, "100", env.Value)
+
+		/* TODO
```
errantepiphany: @spolti , @tchughesiv When I run the unit test, the actual value has somehow become html-escaped. I tried to un-escape it using `html.UnescapeString`, to no avail. I'm pretty stumped here why it's doing that. Help?

errantepiphany: The lines got messed up since I did a force push... Just look for the TODO and see that comment block, please. :)

PR #356, opened 2020-02-05
[BAPL-1507] Add to OpenShift images property for JVM max memory ratio conf

swati-kale:
Corresponding UI changes.

Signed-off-by: Swati Kale <swkale@redhat.com>

### DIFF
--- a/deploy/ui/form.json
+++ b/deploy/ui/form.json
@@ -578,6 +578,146 @@
               ]
             },
             {
+              "label": "Enable JVM configuration",
+              "type": "checkbox",
+              "required": false,
+              "description": "When set to true will configure the RHPAM Workbench with JVM configuration, if no configuration is added, the defaults will be used.",
+              "jsonPath": "$.spec.objects.console.jvm.enableConfiguration",
+              "originalJsonPath": "$.spec.objects.console.jvm.enableConfiguration",
+              "fields": [
+                {
+                  "label": "Enable JVM configuration",
+                  "type": "fieldGroup",
+                  "jsonPath": "$.spec.objects.console.jvm",
+                  "originalJsonPath": "$.spec.objects.console.jvm",
+                  "displayWhen": "true",
+                  "visible": false,
+                  "fields": [
+                    {
+                      "label": "Java Opts append",
+                      "type": "text",
+                      "required": false,
+                      "description": "User specified Java options to be appended to generated options in JAVA_OPTS. e.g. '-Dsome.property=foo'.",
+                      "jsonPath": "$.spec.objects.console.jvm.javaOptsAppend",
+                      "originalJsonPath": "$.spec.objects.console.jvm.javaOptsAppend"
+                    },
+                    {
+                      "label": "Java max memory ratio",
+                      "type": "integer",
+                      "required": false,
+                      "description": "Is used when no '-Xmx' option is given in JAVA_OPTS. This is used to calculate a default maximal heap memory based on a containers restriction. If used in a container without any memory constraints for the container then this option has no effect. If there is a memory constraint then '-Xmx' is set to a ratio of the container available memory as set here. The default is '50' which means 50% of the available memory is used as an upper boundary. You can skip this mechanism by setting this value to '0' in which case no '-Xmx' option is added.",
+                      "jsonPath": "$.spec.objects.console.jvm.javaMaxMemRatio",
+                      "originalJsonPath": "$.spec.objects.console.jvm.javaMaxMemRatio"
+                    },
+                    {
+                      "label": "Java initial memory ratio",
+                      "type": "integer",
+                      "required": false,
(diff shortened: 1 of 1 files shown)


swati-kale:
> I'm guessing this PR will have no impact without a packr file

The Packr files didn not generate at console-form make.

deploy/ui/form.json
```diff
@@ -583,6 +583,146 @@
                 }
               ]
             },
+            {
+              "label": "Enable JVM configuration",
+              "type": "checkbox",
+              "required": false,
+              "description": "When set to true will configure the RHPAM Workbench with JVM configuration, if no configuration is added, the defaults will be used.",
+              "jsonPath": "$.spec.objects.console.jvm.enableConfiguration",
+              "originalJsonPath": "$.spec.objects.console.jvm.enableConfiguration",
```
errantepiphany: @swati-kale There is no corresponding "enableConfiguration" boolean property in the crd:
https://github.com/kiegroup/kie-cloud-operator/blob/master/deploy/crds/kieapp.crd.yaml#L188
@tchughesiv When you submitted a PR to my branch, you had removed this property.
This is a problem, right? It seems like it might need to exist in just for UI purposes, even though is isn't a "real" configuration item that needs to be passed to the container.

deploy/ui/form.json
```diff
@@ -1094,6 +1234,146 @@
                     }
                   ]
                 },
+                {
+                  "label": "Enable JVM configuration",
+                  "type": "checkbox",
+                  "required": false,
+                  "description": "When set to true will configure the KIE Server with JVM configuration, if no configuration is added, the defaults will be used.",
+                  "jsonPath": "$.spec.objects.servers[*].jvm.enableConfiguration",
+                  "originalJsonPath": "$.spec.objects.servers[*].jvm.enableConfiguration",
```
errantepiphany: @swati-kale , @tchughesiv Same concern as above.
https://github.com/kiegroup/kie-cloud-operator/blob/master/deploy/crds/kieapp.crd.yaml#L566

deploy/ui/form.json
swati-kale: @tchughesiv , 'enableConfiguration' is needed for UI checkbox, and even though it will be a par of CR yaml, it will be ignored by operator ,  right?

errantepiphany: @tchughesiv I feel the checkbox is important to have, as there are many jvm properties, and in most cases, users won't care about them. So it should be unchecked by default, effectively hiding the jvm configuration block. A user would have to click the checkbox to show that section. If we don't do this, we would clutter the UI with too many things that again, most users don't need to see (but some would for some use cases).

errantepiphany: I can do the same as before and add the property, but if you could run the code generation (since I'm still having sdk trouble), I would appreciate it. Then this PR can be (rebased and) merged.

errantepiphany: @tchughesiv @swati-kale I've submitted this PR. Tommy, might you be able to re-run the generation on it please, and submit a PR to my branch so I can update this PR?

bmozaffa: @errantepiphany as Swati said, if this is purely needed for the UI at this point, there is no need to add the property. It can just be generated by the UI in the yaml and ignored by the operator. If we supported updates, that would be a different story... same would apply down the line if operand forms require the property.

errantepiphany: @bmozaffa So this PR is unnecessary? https://github.com/kiegroup/kie-cloud-operator/pull/359/files

bmozaffa: I think unnecessary from the narrow pov of this requirement

errantepiphany: @bmozaffa But doesn't `$.spec.objects.console.jvm.enableConfiguration` need to exist so that the `originalJsonPat` line in form.json doesn't blow up? (I'm exposing my ignorance here...)

swati-kale: @errantepiphany , form.json is just for UI reference, and path is for creation of yaml. so will not be a problem . But in case of some fields like dropdown it may need originalpath to refer the enum corresponding to select values. So no issue in this case.

errantepiphany: @swati-kale Okay, now I have a better understanding, thank you. Sooo... we should be able to merge this UI PR, yes?

bmozaffa: To add a bit more context, the CRD is just there to:

1- Enforce what shows up in terms of required fields. We're talking about optional fields, so the CRD has no impact with that regard. In Kubernetes, you can always have extra fields that are not in the CRD, and they will never be disallowed.
2- Document what input should look like, so users know what fields to insert. In this case, the user never provides enableConfiguration in a yaml, because nobody will care about it or use it. If our Operator UI had edit functionality, the case would be different. We'd want the user to provide it even when inputting yaml, so if the environment is later edited through the form, it is rendered correctly. However, we have no plans to provide edit functionality in this UI.

That's why it doesn't need to go into the CRD

bmozaffa: @swati-kale This PR is a "no-op". It doesn't do anything. The operator binary reads form.json from the packr file, which is not changed in the PR.

swati-kale: @bmozaffa The packr file should generate automatically , for this change it didn't generate.  What could be the cause.

bmozaffa: Sorry for the confusion, I forgot how this worked. Whereas in the case of console-cr-form, changes lead to a packr file that is committed to the git repo and passed on to this project, in the case of the operator, the packr file is generated as needed at build and distribution time and never checked in... therefore, the packr file I was looking for is in the [.gitignore](https://github.com/kiegroup/kie-cloud-operator/blob/master/.gitignore#L6) and will not be needed here. Please exclude all other files and only keep form.json in this PR, and let's merge it.

PR #361, opened 2020-02-11
[KIECLOUD-358] sha retrieval, disconnectedMode, and csv version changes

tchughesiv:
`spec.useImageTags` is `false` by default. Setting to true will disable SHA digest utilization in favor of image tags for product versions 7.7.0 and greater. 


deploy/ui/form.json
```diff
@@ -47,6 +47,14 @@
             }
           ]
         },
+        {
+          "label": "Disable Disconnected Mode",
```
bmozaffa: Shouldn't we just say "Use Image Tags" here and explain the consequences in the description as you do?

tools/csv-gen/csv-gen.go
```diff
@@ -239,6 +243,12 @@ func main() {
 						Path:         "upgrades.minor",
 						XDescriptors: []string{"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"},
 					},
+					{
+						Description:  "Set true to disable disconnected install mode and, instead, enable image tags. Image Tags are disabled by default. This will leverage image tags instead of the image SHA digests.",
+						DisplayName:  "Disable Disconnected Mode",
```
bmozaffa: Same question here

deploy/ui/form.json
tchughesiv: done

tools/csv-gen/csv-gen.go
tchughesiv: done

PR #385, opened 2020-03-10
Add OCP web console enrichments (ocp4.3 or after)

myeung18:
Signed-off-by: Marco Yeung <myeung@redhat.com>


pkg/controller/initializer.go
```diff
@@ -13,6 +14,7 @@ func init() {
 	addManager := func(mgr manager.Manager) error {
 		k8sService := GetInstance(mgr)
 		reconciler := kieapp.Reconciler{Service: &k8sService}
+		webconsole.LoadWebConsoleEnrichmentUI()
```
bmozaffa: It seems to me that this is similar to creating our operator UI. We do that [here](https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/controller/kieapp/kieapp_controller.go#L58-L60), so I would say this line would go right next to it. @tchughesiv , wdyt?

pkg/controller/kieapp/webconsole/webconsole.go
```diff
@@ -0,0 +1,102 @@
+package webconsole
```
bmozaffa: Maybe just call it olm?

pkg/controller/kieapp/webconsole/webconsole.go
```diff
@@ -0,0 +1,102 @@
...
+	files, err := loadFilesOnlyWithBox("consoleyamlsamples", "noop", "noop")
+	if err != nil {
+		log.Error(err, "yaml stings not loaded")
+	} else {
+		_, err = applyMultipleWebConsoleYamls(files)
+		if err != nil {
+			log.Error(err, "web console yamlsamples not successfully applied")
+		}
+	}
+	files, err = loadFilesOnlyWithBox("webconsole", "noop", "noop")
```
bmozaffa: This operator only needs sample yamls in this code. The only other thing is the link, but it won't be based on a yaml and will have to be separately figured out based on the deployed object, so please take out the other stuff from here and just focus on creating the yaml samples.

bmozaffa: Given the above, chance the function name to something like "CreateSampleYAMLs"

pkg/controller/kieapp/webconsole/webconsole.go
```diff
@@ -0,0 +1,102 @@
...
+		log.Error(err, "yaml stings not loaded")
+	} else {
+		_, err = applyMultipleWebConsoleYamls(files)
+		if err != nil {
+			log.Error(err, "web console enrichments not successfully applied")
+		}
+	}
+}
+
+func applyMultipleWebConsoleYamls(yamlString []string) (map[int]string, error) {
```
bmozaffa: I'm not sure why we need all of this. It's 1 line to create an object, and an extra 3-4 lines to handle the error. For example, see [here](https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/controller/kieapp/kieapp_controller.go#L682-L686).

You just need to create an object type based on the existing yaml structure that the code can then create.

ruromero:
Consider depending on this PR https://github.com/kiegroup/kie-cloud-operator/pull/386

pkg/controller/kieapp/webconsole/webconsole.go
myeung18: probably not as all these customizations are meant for OCP Container platform and independent of the olm.

pkg/controller/kieapp/webconsole/webconsole.go
myeung18: modified.

pkg/controller/initializer.go
myeung18: moved. 
it looks like a proper place to put the yaml sample creation method call.

pkg/controller/kieapp/webconsole/webconsole.go
myeung18: These are all here is meant to deal with a bunch of yaml files passed in, and we don't know exactly what type they are (unless we read the file and get the kind), they could be yaml sample, consolelink, consoleexternaloglink etc.
All these are used to find the object type and then delegate to the right creator to create.

bmozaffa: Yes, I am not aware of any different type of yaml that needs to be handled in this code, or a situation where the caller is passing in a yaml and doesn't know what it is. Let's make this quick and simple, to handle the only types we need to create here, which is CR samples.

pkg/controller/kieapp/webconsole/webconsole.go
bmozaffa: Sure, I guess webconsole is fine

tchughesiv:
please rebase

pkg/controller/kieapp/webconsole/webconsole.go
myeung18: That means we don't need the factor and creator. Modified.

pkg/controller/kieapp/webconsole/webconsole.go
```diff
@@ -0,0 +1,136 @@
...
+type CustomResourceDefinition struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+}
+
+var log = logs.GetLogger("kieapp.webconsole")
+
+func CreateSampleYAMLs() {
+	log.Info("Loading Console yaml samples.")
+	files, err := loadFilesOnlyWithBox("consoleyamlsamples")
```
bmozaffa: This should use actual files in `deploy/crs`. I'm not sure what this currently does or denotes.

pkg/controller/kieapp/webconsole/webconsole.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			resMap[idx] = err.Error()
+		} else {
+			resMap[idx] = "Applied"
+		}
+	}
+	return resMap, nil
+}
+
+func create(yamlStr string) (bool, error) {
```
bmozaffa: There is a whole lot of complexity here, and I'm not sure why. The operator already has a `writer` that is used throughout the project to create resources.

pkg/controller/kieapp/webconsole/webconsole.go
myeung18: currently for testing, now moved to crs folder.

pkg/controller/kieapp/webconsole/webconsole.go
myeung18: Should've used the reconciler from controller. Modified.
now only webconsole.go is left.

deploy/crs/kieapp_rhdm_trial.yaml
```diff
@@ -2,5 +2,10 @@ apiVersion: app.kiegroup.org/v2
 kind: KieApp
 metadata:
   name: rhdm-trial
+  annotations:
+    consolekind: ConsoleYAMLSample
+    consolename: kieapp-rhdm-trial
+    consoledesc: KieApp rhdm trial
+    consoletitle: KieApp rhdm trial
```
bmozaffa: Please use camel case, e.g., consoleTitle

deploy/crs/kieapp_rhdm_trial.yaml
```diff
@@ -2,5 +2,10 @@ apiVersion: app.kiegroup.org/v2
 kind: KieApp
 metadata:
   name: rhdm-trial
+  annotations:
+    consolekind: ConsoleYAMLSample
```
bmozaffa: Why is this necessary?

pkg/controller/kieapp/webconsole.go
```diff
@@ -0,0 +1,141 @@
...
+	TargetNamespace   = "openshift-console"
+	ConsoleYAMLSample = "ConsoleYAMLSample"
+)
+
+type CustomResourceDefinition struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+}
+
+func CreateConsoleYAMLSamples(r *Reconciler) {
```
bmozaffa: This should be package private, so start with lowercase: `create`

pkg/controller/kieapp/webconsole.go
```diff
@@ -0,0 +1,141 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	TargetNamespace   = "openshift-console"
+	ConsoleYAMLSample = "ConsoleYAMLSample"
+)
+
+type CustomResourceDefinition struct {
```
bmozaffa: Shouldn't need this

pkg/controller/kieapp/webconsole.go
```diff
@@ -0,0 +1,141 @@
...
+		} else {
+			resMap[idx] = "Applied"
+		}
+	}
+	return resMap, nil
+}
+
+func create(yamlStr string, r *Reconciler) (bool, error) {
+	obj := &CustomResourceDefinition{}
+	err := yaml.Unmarshal([]byte(yamlStr), obj)
```
bmozaffa: Our unit tests or somewhere should have the right way of reading our CR files

pkg/controller/kieapp/webconsole.go
```diff
@@ -0,0 +1,141 @@
...
+				APIVersion: obj.TypeMeta.APIVersion,
+				Kind:       obj.TypeMeta.Kind,
+			},
+			Title:       consolev1.ConsoleYAMLSampleTitle(title),
+			Description: consolev1.ConsoleYAMLSampleDescription(desc),
+			YAML:        consolev1.ConsoleYAMLSampleYAML(yamlStr),
+			Snippet:     snippet,
+		},
+	}
+	err = r.Service.Create(context.TODO(), yamlSample)
```
bmozaffa: Now thinking about what can be generic, this line won't be portable. I think we should make a simple util function that takes the CR and returns a `ConsoleYAMLSample`, and let the operator create it.

deploy/crs/kieapp_rhdm_trial.yaml
myeung18: It is not needed. However, doing gives us flexibility to selective apply certain CRs (ie:apply only those have this 'consoleKind'), instead of apply the CRs in the entire folder.

bmozaffa: The caller will be passing a yaml to create a console yaml sample. They will want it created and we shouldn't be second-guessing their call. In fact, we should use defaults if no annotations exist at all.

deploy/crs/kieapp_rhdm_trial.yaml
myeung18: modified

pkg/controller/kieapp/webconsole.go
myeung18: modified

pkg/controller/kieapp/webconsole.go
myeung18: removed

pkg/controller/kieapp/webconsole.go
myeung18: modified

pkg/controller/kieapp/webconsole.go
myeung18: modified

pkg/controller/kieapp/shared/utils.go
```diff
@@ -174,3 +180,52 @@ func GetNamespacedName(object metav1.Object) types.NamespacedName {
...
 }
+
+const (
+	TargetNamespace   = "openshift-console"
+	ConsoleYAMLSample = "ConsoleYAMLSample"
+)
+
+func GetConsoleYAMLSample(res resource.KubernetesResource) (*consolev1.ConsoleYAMLSample, error) {
+	kind := res.GetAnnotations()["consoleKind"]
+	if kind != ConsoleYAMLSample {
```
bmozaffa: As mentioned, we should remove this. All the annotations should be optional.

pkg/controller/kieapp/shared/utils.go
```diff
@@ -174,3 +180,52 @@ func GetNamespacedName(object metav1.Object) types.NamespacedName {
...
+		if err != nil {
+			return nil, errors.New("Unable to parse snippet to boolean.")
+		}
+		snippet = tmp
+	}
+	title, _ := res.GetAnnotations()["consoleTitle"]
+	desc, _ := res.GetAnnotations()["consoleDesc"]
+	name, _ := res.GetAnnotations()["consoleName"]
+
+	res.SetAnnotations(make(map[string]string))
```
bmozaffa: They may have their own annotations that need to be in the CR. Only remove our 3 known annotations.

pkg/controller/kieapp/shared/utils.go
```diff
@@ -174,3 +180,51 @@ func GetNamespacedName(object metav1.Object) types.NamespacedName {
 		Namespace: object.GetNamespace(),
 	}
 }
+
+func GetConsoleYAMLSample(res resource.KubernetesResource) (*consolev1.ConsoleYAMLSample, error) {
+	snippetStr := res.GetAnnotations()["consoleSnippet"]
```
bmozaffa: I'm not clear on what this does and what purpose it serves. So it's ok to leave it out, but if the operator passes in a yaml with consoleSnippet set to false, we return an error instead of a ConsoleYAMLSample? Not sure that's necessary, or why an operator would pass it in, if they've marked as false / not a snippet. I would just remove this whole part.

pkg/controller/kieapp/shared/utils.go
```diff
@@ -174,3 +180,51 @@ func GetNamespacedName(object metav1.Object) types.NamespacedName {
...
+	if snippetStr != "" {
+		tmp, err := strconv.ParseBool(snippetStr)
+		if err != nil {
+			return nil, errors.New("Unable to parse snippet to boolean.")
+		}
+		snippet = tmp
+	}
+
+	title, _ := res.GetAnnotations()["consoleTitle"]
+	desc, _ := res.GetAnnotations()["consoleDesc"]
```
bmozaffa: I haven't seen how things should look in the web console... so I take it that title and description is optional? Fine if it is, just checking.

pkg/controller/kieapp/shared/utils.go
```diff
@@ -174,3 +180,51 @@ func GetNamespacedName(object metav1.Object) types.NamespacedName {
...
+	name, _ := res.GetAnnotations()["consoleName"]
+	if name == "" {
+		name = res.GetName() + "-yamlsample"
+	}
+
+	curannos := res.GetAnnotations()
```
bmozaffa: How about giving this a better name (maybe just `annotations`?), moving this further up, and using it to retrieve the values too?

pkg/controller/kieapp/shared/utils.go
```diff
@@ -174,3 +180,51 @@ func GetNamespacedName(object metav1.Object) types.NamespacedName {
...
+				Kind:       res.GetObjectKind().GroupVersionKind().Kind,
+			},
+			Title:       consolev1.ConsoleYAMLSampleTitle(title),
+			Description: consolev1.ConsoleYAMLSampleDescription(desc),
+			YAML:        consolev1.ConsoleYAMLSampleYAML(string(data)),
+			Snippet:     snippet,
+		},
+	}
+	return yamlSample, nil
+}
```
bmozaffa: I think this looks good otherwise. Please find a good folder and package name for it in operator-utils and open a PR there. Also create a unit test for it there, and let's prepare to merge it.

pkg/controller/kieapp/shared/utils.go
myeung18: Since now we by default try our best to create the ConsoleYAMLSample, this error return is not necessary, and we should just give a default value (false: not a snippet).   But we should still have consoleSnippet to allow user to turn the entire yaml as snippet: set to true.
I'd do like this:
	var snippet bool = false
          .....
		tmp, err := strconv.ParseBool(snippetStr)
		if err == nil {
			snippet = tmp
		}

pkg/controller/kieapp/shared/utils.go
myeung18: They are optional.

pkg/controller/kieapp/shared/utils.go
myeung18: modified

pkg/controller/kieapp/shared/utils.go
myeung18: will do.  first, will make this kie-cloud pr working first.

pkg/controller/kieapp/shared/utils.go
bmozaffa: I don't understand!

myeung18: by default this yaml file is not a snippet.  If user wants to mark it as snippet, he gives the consoleSnippet: true. But if the string value has typo: not parsable, we keep it as false.

bmozaffa: oh, I misunderstood this whole thing. What is a snippet? That means the yaml is not your whole CR? Is there a difference in behavior, perhaps in that it doesn't completely replace your yaml when clicked?

myeung18: Correct, different behavior, doesn't completely replace when clicked.
another tab will be created and named "Snippet", normal Yamls have "Samples" tab.

pkg/controller/kieapp/shared/utils.go
```diff
@@ -174,3 +180,50 @@ func GetNamespacedName(object metav1.Object) types.NamespacedName {
...
 	}
 }
+
+func GetConsoleYAMLSample(res resource.KubernetesResource) (*consolev1.ConsoleYAMLSample, error) {
+	annotations := res.GetAnnotations()
+	snippetStr := annotations["consoleSnippet"]
+	var snippet bool = false
+	if snippetStr != "" {
+		tmp, err := strconv.ParseBool(snippetStr)
+		if err == nil {
```
bmozaffa: Nothing wrong with this, but you can also do:

```
if tmp, err := strconv.ParseBool(snippetStr); err == nil {
   snippet = tmp
}
```

pkg/controller/kieapp/webconsole/webconsole.go
myeung18: done

deploy/crs/kieapp_rhdm_authoring.yaml
```diff
@@ -2,5 +2,6 @@ apiVersion: app.kiegroup.org/v2
 kind: KieApp
 metadata:
   name: rhdm-authoring
+  annotations:
```
bmozaffa: Why this? Did you find annotations needed to be there or something?

pkg/controller/kieapp/webconsole.go
```diff
@@ -0,0 +1,99 @@
...
+		return false, err
+	}
+	err = r.Service.Create(context.TODO(), yamlSample)
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func loadFiles(boxname string, path ...string) ([]string, []string, error) {
```
bmozaffa: Why is a boxname needed? Why not just hardcode to something on line 75?

deploy/crs/kieapp_rhdm_authoring.yaml
myeung18: cleaned, I need more vigilance.

pkg/controller/kieapp/webconsole.go
myeung18: that way looks better, modified.

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/heroku/docker-registry-client v0.0.0-20190909225348-afc9e1acc3d5
 	github.com/imdario/mergo v0.3.8
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
-	github.com/openshift/client-go v0.0.0-20200116152001-92a2713fa240
+	github.com/openshift/client-go v3.9.0+incompatible
```
bmozaffa: This version name is fishy! Is this right? No adverse effects? @tchughesiv ?

pkg/controller/kieapp/webconsole.go
```diff
@@ -0,0 +1,93 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"github.com/RHsyseng/operator-utils/pkg/utils/openshift"
+	"github.com/ghodss/yaml"
+	"github.com/gobuffalo/packr/v2"
+	api "github.com/kiegroup/kie-cloud-operator/pkg/apis/app/v2"
+)
+
+func CreateConsoleYAMLSamples(r *Reconciler) {
```
bmozaffa: I think this file will look much better if you combine these 4 into one function.

myeung18: modified

pkg/controller/kieapp/webconsole.go
```diff
@@ -0,0 +1,60 @@
...
+		resMap[idx] = "Applied"
+	}
+
+	for k, v := range resMap {
+		log.Infof("yaml file: %s %s ", names[k], v)
+	}
+}
+
+//used by packr
+func convertYamlToBinary() {
```
bmozaffa: Is this old / should be removed?

myeung18: redundant.

go.mod
tchughesiv: there should be no issues with this as the important line in go.mod for this library is this one in the replace section - https://github.com/kiegroup/kie-cloud-operator/blob/ad198e7e68bafeb5e477dbfe1e45df40cbb35520/go.mod#L48

which forces this in go.sum - https://github.com/kiegroup/kie-cloud-operator/blob/ad198e7e68bafeb5e477dbfe1e45df40cbb35520/go.sum#L668

that said, i'm not sure why this line is changing with this PR... maybe carryover from he operator-utils library change?

bmozaffa: Any thoughts, Marco?

myeung18: I also think the version at the replace section will take the precedence over that version at the require section. 
That line seem to be generated by go mod as we didn't specify the exact version when operator-utils was converted to use go mod from dep. Now it is carried forward.
So, I think we should be ok.

Btw, why do we use a lot of replace sections given that we know exactly what version we want?
If an exact version was listed in operator-utils' require section, could we have less this  type of version concern?

bmozaffa: Good question, Marco. I'm not sure and maybe Tommy is better placed to answer this question. I suppose we want to be thorough and avoid this project breaking in case operator-utils uses a different version. In other words, avoid tight coupling to that project.

PR #390, opened 2020-03-16
[RHPAM-2684] Adapt to operator-sdk 0.14.1 scaffolding

ruromero:


vendor/modules.txt
```diff
@@ -196,7 +196,6 @@ github.com/operator-framework/operator-sdk/pkg/kube-metrics
 github.com/operator-framework/operator-sdk/pkg/leader
 github.com/operator-framework/operator-sdk/pkg/log/zap
 github.com/operator-framework/operator-sdk/pkg/metrics
-github.com/operator-framework/operator-sdk/pkg/restmapper
```
bmozaffa: Let's figure out why the imports moved. I want to avoid moving them back and forth with PRs, so if there is `go fmt` or similar is sorting them, let's go with that. Otherwise, avoid rearranging.

ruromero: It appeared when I used sdk 0.14.1

bmozaffa: I'm talking about lines going up and down... not imports being added/removed.

bmozaffa: I should have made this comment further down... for example, look at lines 406,-407 versus 409 and 415

bmozaffa: Line movements seems to be generated this way with new go versions.

PR #413, opened 2020-04-09
[KIECLOUD-382] modify brew digest retrieval to use stage registry

tchughesiv:
/hold
/assign @spolti 

Maybe using stage registry to populate digests accomplishes our needs for now? At least until ART automation is finished which will handle a lot of that for us?

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,6 @@ CSV Generation
 make csv
 
 # OR
-# w/ sha lookup/replacement against registry.redhat.io
-DIGESTS=true USER_TOKEN="<username>:<password>" make csv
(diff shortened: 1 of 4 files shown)


tchughesiv:
thinking about this more... i think we may have to split this up so that only rhpam/dm images are pulled from stage, but the rest from registry.redhat.io. 

I'm concerned that some ancillary image we rely on has a digest set in stage but hasn't yet been published when we release. That could be bad.

tchughesiv:
ok @spolti  ... i've worked something out with this. for current product release images, in this case 7.8.0, digests will be pulled from the stage registry. For all other images and product versions, we'll pull digest from registry.redhat.io.

PR #418, opened 2020-04-14
VersionCheck for YAML Samples

myeung18:
Signed-off-by: Marco Yeung <myeung@redhat.com>

add logic for OCP version check before applying YAML Samples

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/RHsyseng/console-cr-form v0.0.0-20200414161125-135bc9b52976
-	github.com/RHsyseng/operator-utils v0.0.0-20200407154515-94a3ad1cdcfc
+	github.com/RHsyseng/operator-utils v0.0.0-20200414154239-8ab0e9a7d0f7
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/coreos/prometheus-operator v0.35.1
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
@@ -37,7 +37,7 @@ require (
 replace (
 	// Pin RHsyseng library versions
 	github.com/RHsyseng/console-cr-form => github.com/RHsyseng/console-cr-form v0.0.0-20200414161125-135bc9b52976
-	github.com/RHsyseng/operator-utils => github.com/RHsyseng/operator-utils v0.0.0-20200407154515-94a3ad1cdcfc
+	github.com/RHsyseng/operator-utils => github.com/RHsyseng/operator-utils v0.0.0-20200414154239-8ab0e9a7d0f7
 
 	github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309 // Required by Helm
 	github.com/gobuffalo/packr/v2 => github.com/gobuffalo/packr/v2 v2.7.1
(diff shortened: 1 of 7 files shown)


go.sum
```diff
@@ -52,10 +52,17 @@ github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbt
 github.com/PuerkitoBio/urlesc v0.0.0-20160726150825-5bd2802263f2/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 h1:d+Bc7a5rLufV/sSk/8dngufqelfh6jnri85riMAaF/M=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
+<<<<<<< HEAD
```
tchughesiv, line 55: looks like maybe some leftover lines from rebase?

pkg/controller/initializer.go
```diff
@@ -37,7 +37,12 @@ func init() {
 				log.Warn("OpenShift version could not be determined.")
 			}
 		}
-		kieapp.CreateConsoleYAMLSamples(&reconciler)
+		res, err := openshift.CompareOpenShiftVersion(nil, "4.3")
```
tchughesiv, line 40: what does this function do?

myeung18: this func makes sure the current platform is Openshift, and compare its version with the version passed in, then result the comparison result.
Here the code proceed to create the YAML samples only when this is OCP, and the version is >= 4.3.

go.sum
myeung18: should be fixed now.

go.mod
```diff
@@ -36,8 +35,8 @@ require (
 
 replace (
 	// Pin RHsyseng library versions
-	github.com/RHsyseng/console-cr-form => github.com/RHsyseng/console-cr-form v0.0.0-20200414161125-135bc9b52976
-	github.com/RHsyseng/operator-utils => github.com/RHsyseng/operator-utils v0.0.0-20200407154515-94a3ad1cdcfc
+	github.com/RHsyseng/console-cr-form => github.com/RHsyseng/console-cr-form v0.0.0-20200129200812-dc119cb0bd4d
```
tchughesiv, line 38: why is this being reverted to an older version? is this module related to this PR somehow?

pkg/controller/initializer.go
tchughesiv: hmm, seems like this is duplicate code to what's being done above it. since the openshift version is already known at this point in the flow, couldn't we move this up into the `info.IsOpenShift()` if statement and do a simple `>= "4.3"`?

myeung18: modified.

pkg/controller/initializer.go
```diff
@@ -36,8 +36,12 @@ func init() {
 			} else {
 				log.Warn("OpenShift version could not be determined.")
 			}
+			if mappedVersion.Version < "4.3" {
```
bmozaffa, line 39: @myeung18 @tchughesiv this looks like a string compare, so I wonder what the repercussions are. For example, I imagine 4.11 < 4.3

tchughesiv: i just checked this with a unit test and it won't work... considers 4.3 to be greater than 4.11. That's my fault... will have to make a few changes in code besides this one.

myeung18: I am also aware the bug here. I am working on a fix. I will show probably tmr..

PR #419, opened 2020-04-15
Switch to semver for product & ocp version checks

tchughesiv:
- `golang.org/x/mod/semver` for ocp and product version checks
 - pim enabled only for product version 7.8+
 - ocp 4.4 added
 - oauth image tags changed to use `v4.x` format
 - ConsoleLink and `inject-trusted-cabundle` ConfigMap only supported in OCP 4.2+
 - ConsoleYAMLSamples only supported in OCP 4.3+

### DIFF
--- a/pkg/controller/initializer.go
+++ b/pkg/controller/initializer.go
@@ -6,7 +6,10 @@ import (
 	"github.com/RHsyseng/operator-utils/pkg/logs"
 	"github.com/RHsyseng/operator-utils/pkg/utils/kubernetes"
 	"github.com/RHsyseng/operator-utils/pkg/utils/openshift"
+	"github.com/blang/semver"
 	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp"
+	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp/constants"
+	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp/shared"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
@@ -25,20 +28,18 @@ func init() {
 			mappedVersion := openshift.MapKnownVersion(info)
 			if mappedVersion.Version != "" {
 				log.Info(fmt.Sprintf("OpenShift Version: %s", mappedVersion.Version))
-				reconciler.OcpVersion = mappedVersion.Version
-				reconciler.OcpVersionMajor = mappedVersion.MajorVersion()
-				reconciler.OcpVersionMinor = mappedVersion.MinorVersion()
-				/* ?? warning if ocp version isn't in SupportedOcpVersions slice ??
-				if _, ok := shared.Find(constants.SupportedOcpVersions, reconciler.OcpVersion); !ok {
+				if _, ok := shared.Find(constants.SupportedOcpVersions, mappedVersion.Version); !ok {
 					log.Warn("OpenShift version not supported.")
 				}
-				*/
+				if v, err := semver.New(mappedVersion.Version + ".0"); err == nil {
+					reconciler.OcpVersion = v
+				} else {
+					log.Warn("OpenShift version could not be parsed.")
+				}
 			} else {
 				log.Warn("OpenShift version could not be determined.")
 			}
-			if mappedVersion.Version < "4.3" {
-				log.Warn("Console YAML sample is not added, curVer:", mappedVersion.Version)
-			} else {
+			if reconciler.OcpVersion == nil || reconciler.OcpVersion.GE(semver.MustParse("4.3.0")) {
 				kieapp.CreateConsoleYAMLSamples(&reconciler)
 			}
 		}
--- a/pkg/controller/initializer.go
+++ b/pkg/controller/initializer.go
@@ -6,10 +6,10 @@ import (
 	"github.com/RHsyseng/operator-utils/pkg/logs"
 	"github.com/RHsyseng/operator-utils/pkg/utils/kubernetes"
 	"github.com/RHsyseng/operator-utils/pkg/utils/openshift"
-	"github.com/blang/semver"
 	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp"
 	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp/constants"
 	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp/shared"
+	"golang.org/x/mod/semver"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
@@ -27,21 +27,17 @@ func init() {
 		if info.IsOpenShift() {
 			mappedVersion := openshift.MapKnownVersion(info)
 			if mappedVersion.Version != "" {
-				log.Info(fmt.Sprintf("OpenShift Version: %s", mappedVersion.Version))
 				if _, ok := shared.Find(constants.SupportedOcpVersions, mappedVersion.Version); !ok {
 					log.Warn("OpenShift version not supported.")
 				}
-				if v, err := semver.New(mappedVersion.Version + ".0"); err == nil {
-					reconciler.OcpVersion = v
-				} else {
-					log.Warn("OpenShift version could not be parsed.")
-				}
+				reconciler.OcpVersion = semver.MajorMinor("v" + mappedVersion.Version)
+				log.Info(fmt.Sprintf("OpenShift Version: %s", reconciler.OcpVersion))
 			} else {
 				log.Warn("OpenShift version could not be determined.")
 			}
-			if reconciler.OcpVersion == nil || reconciler.OcpVersion.GE(semver.MustParse("4.3.0")) {
-				kieapp.CreateConsoleYAMLSamples(&reconciler)
-			}
+		}
+		if semver.Compare(reconciler.OcpVersion, "v4.3") >= 0 || reconciler.OcpVersion == "" {
+			kieapp.CreateConsoleYAMLSamples(&reconciler)
 		}
 		return kieapp.Add(mgr, &reconciler)
 	}
--- a/deploy/catalog_resources/community/7.8/image-references
+++ b/deploy/catalog_resources/community/7.8/image-references
@@ -112,15 +112,19 @@ spec:
     name: golang-github-openshift-oauth-proxy-container
   - from:
       kind: DockerImage
-      name: registry.redhat.io/openshift4/ose-oauth-proxy:4.3
+      name: registry.redhat.io/openshift4/ose-oauth-proxy:v4.4
     name: golang-github-openshift-oauth-proxy-container
   - from:
       kind: DockerImage
-      name: registry.redhat.io/openshift4/ose-oauth-proxy:4.2
+      name: registry.redhat.io/openshift4/ose-oauth-proxy:v4.3
     name: golang-github-openshift-oauth-proxy-container
   - from:
       kind: DockerImage
-      name: registry.redhat.io/openshift4/ose-oauth-proxy:4.1
+      name: registry.redhat.io/openshift4/ose-oauth-proxy:v4.2
+    name: golang-github-openshift-oauth-proxy-container
+  - from:
+      kind: DockerImage
+      name: registry.redhat.io/openshift4/ose-oauth-proxy:v4.1
     name: golang-github-openshift-oauth-proxy-container
   - from:
       kind: DockerImage
--- a/deploy/catalog_resources/community/7.8/kiecloud-operator.7.8.0.clusterserviceversion.yaml
+++ b/deploy/catalog_resources/community/7.8/kiecloud-operator.7.8.0.clusterserviceversion.yaml
@@ -7,7 +7,7 @@ metadata:
     categories: Integration & Delivery
     certified: "false"
     containerImage: quay.io/kiegroup/kie-cloud-operator:7.8.0
-    createdAt: "2020-04-30 11:12:47"
+    createdAt: "2020-05-01 11:17:24"
     description: Deploys and manages Red Hat Process Automation Manager and Red Hat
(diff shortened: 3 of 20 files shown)


myeung18:
> @myeung18 @bmozaffa
> Would something like this work? Ideally though, I'd like to see OpenShiftVersion become something we could use publicly instead of having to recreate anything in kie operator.
> https://github.com/RHsyseng/operator-utils/blob/a0a604e9efc77f7369755629e2215fd0cc2ede7c/internal/platform/types.go#L45-L102
> 
> Signed-off-by: tchughesiv [tchughesiv@gmail.com](mailto:tchughesiv@gmail.com)



> @myeung18 @bmozaffa
> Would something like this work? Ideally though, I'd like to see OpenShiftVersion become something we could use publicly instead of having to recreate anything in kie operator.
> https://github.com/RHsyseng/operator-utils/blob/a0a604e9efc77f7369755629e2215fd0cc2ede7c/internal/platform/types.go#L45-L102
> 
> Signed-off-by: tchughesiv [tchughesiv@gmail.com](mailto:tchughesiv@gmail.com)

Looks fine for kie-cloud operator due to lack of common Version struct to operator-utils; otherwise, those version comparison could all be in operator-utils. 

But I do would like to modify the CompareOpenShiftVersion to CompareOpenShiftVersion(ver1 string, ver2 string), because it is now doing two things in a single method (check openshift and compare).  If we can make this new func, maybe kie-cloud could also use it?

bmozaffa:
Can/should we use semver package?

If not and if like this, maybe better if OpenShiftVersion struct has the compare operator and you're forced to create one before comparing, so the input format is enforced.

myeung18:
> Can/should we use semver package?
adding semver, the only thing we gain is to enforce the input: not using string?
I see no other benefit.

> 
> If not and if like this, maybe better if OpenShiftVersion struct has the compare operator and you're forced to create one before comparing, so the input format is enforced.

You mean make OpenShiftVersion Comparable?
OpenShiftVersion now does have Compare operator.

New PR#
https://github.com/RHsyseng/operator-utils/pull/58

tchughesiv:
@bmozaffa @myeung18 i've added changes using semver instead, to see how that might work... take a look please

tchughesiv:
maybe we can even offer a function in operator-utils that passes back a semver.Version type

myeung18:
> @bmozaffa @myeung18 i've added changes using semver instead, to see how that might work... take a look please

It does have advanced features and fancy as well.  The version our operator-utils can handle, however, is only major.minor anyway.  A self-made simple comparison handling is enough for now in my opinion.

pkg/controller/initializer.go
```diff
@@ -2,11 +2,15 @@ package controller
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/RHsyseng/operator-utils/pkg/logs"
 	"github.com/RHsyseng/operator-utils/pkg/utils/kubernetes"
 	"github.com/RHsyseng/operator-utils/pkg/utils/openshift"
+	"github.com/blang/semver"
```
bmozaffa, line 10: Did you check out golang.org/x/mod/semver?

myeung18: tend to use string and with this format (vMajor.Minor : v4.3) for data  transmission?

tchughesiv: just tested a few things... appears to require a `v` to be prepended to all versions.

tchughesiv: `v4.3` for example

bmozaffa: So the built-in semver library requires that, and this one doesn't? I imagine we could work around that by adding the "v" at the beginning, no? I remember reading about plans to adopt semver in OCP, or it may have been somewhere else!

I would do a cost/benefit analysis in this case... the "v" at the beginning being a bit of a hack, but on the other hand, will it be a similar library dependency or a "built-in" feature?

tchughesiv: i'd agree that if ocp was going to adopt that version of semver... we should too. will try to find out

tchughesiv: might look something like this ... https://github.com/tchughesiv/kie-cloud-operator/compare/fixVersionCompare...tchughesiv:gSemverNewer

bmozaffa: Looks good to me! What do you think, @myeung18 ?

Curious, does this avoid vendoring? I'm not sure how the golang/x works

tchughesiv: still vendors... at least w/ go 1.12 ... maybe that changes in 1.13?

myeung18: For operator-uilts, I will take string as the type to accept version params. Internally, I will verify these input align with mod.semver format, if good, I will trim that 'v' out, and use the OpenShiftVersion to do the rest: do the comparison?  Since in operator-utils, we are still dealing with OpenShiftVersion, does this sound good?

tchughesiv: any chance we can move OpenShiftVersion type out from internal section so I can use that type in BA operator?

myeung18: @jeremyary will there be any big impact or actually easy to do?

bmozaffa: @myeung18 not sure I follow. I thought the idea was to use semver to do the comparison. If so, why wouldn't you do that in operator-utils?

myeung18: we can use do that, was just asking, since OpenShiftVersion is now Comparable thought I should make use of it.

Comparsion in operator-uilts will be using mod.semver then.

bmozaffa: If we think semver makes sense, we should rethink the design of OpenShiftVersion based on that.

myeung18: then the version string return from operator-utils:PlatformInfo, do we keep the existing one or should we make it consistent with this new format?

jeremyary: I thought SemVer was what I linked in the previous PR as what I wanted to chase for redesign, but you said it had permissions issues?

bmozaffa: I haven't looked at it. Please think about it, try to redesign and let's review when ready.

jeremyary: > I thought SemVer was what I linked in the previous PR as what I wanted to chase for redesign, but you said it had permissions issues?

https://github.com/RHsyseng/operator-utils/pull/53#discussion_r406310699

Regardless, yes, if this works without elevated perms as I was hoping, it garners a redesign that should do away with some of the complication in operator-utils.

myeung18: @tchughesiv @jeremyary @bmozaffa 
New PR#
RHsyseng/operator-utils#58

tchughesiv: @jeremyary @myeung18 i think we're getting closer to something usable with that PR, but can someone help me understand the reasoning behind making OpenShiftVersion an `internal` only package? I'd like to import it but am unable.

jeremyary: Well, operator-utils has grown a lot since this was implemented last June - this was originally a [refactor of a catch-all 'utils' file](https://github.com/RHsyseng/operator-utils/pull/16/files) that was building up and becoming more deserving of its own package. The original intent when worked out between @bmozaffa and myself was utils.go would put forward an easy-to-read set of endpoints from a single import for the user to consume all of our helper methods without having to filter through implementation (and there were already some projects using method signatures from this utils.go file). There may have been something around the Discoverer interface/testability that led in this direction as well but I can't recall and the convo took place outside of PR's and I can't find it now.

Pretty well everything we've added to the project since hasn't followed this dynamic, likely because our understanding of Go paradigms grew along with the codebase - mine for certain... this feature was where I by and large cut my teeth as a first major go contribution.

myeung18: We probably have better use case to make OpenShifftVersion public and act as carrier.
this how it would look like: https://github.com/RHsyseng/operator-utils/pull/59
platform and openshift pkg are now cross referencing, however. Before it was only one way.

myeung18: Just for the Information, a cut and paste of the pkg from internal to utils pkg will also do the job instead of breaking the current cod. Only if we decided to do so.

myeung18: @tchughesiv @jeremyary @bmozaffa

bmozaffa: Was Jeremy planning on taking a second look at the whole approach at a high level?

spolti:
for ocp 3.11 there is a issue regarding the Console Link, seems that ocp 3.11 does not support it.
here is the stacktrace:

```2020-04-27T17:01:28.602-0300	ERROR	controller-runtime.controller	Reconciler error	{"controller": "kieapp-controller", "request": "rhpam-2749/rhpam-authoring-ha", "error": "no matches for kind \"ConsoleLink\" in version \"console.openshift.io/v1\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/github.com/go-logr/zapr/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:153
k8s.io/apimachinery/pkg/util/wait.Until
	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
```

Probably this step needs to be skipped on ocp version < 4.

tchughesiv:
we also need to add a check for which product version we're using related to PIM and the new db config structure. with latest code, when setting to 7.7.0 version, we get following error.
```yaml
  - lastTransitionTime: "2020-04-28T19:56:24Z"
    message: 7.7.0/dbs/servers/h2.yaml does not exist, 'rhpam-trial' KieApp not deployed
    reason: ConfigurationError
    status: "True"
    type: Failed
    version: 7.7.0
```

tchughesiv:
> for ocp 3.11 there is a issue regarding the Console Link, seems that ocp 3.11 does not support it.
> here is the stacktrace:
> 
> ```
> github.com/go-logr/zapr.(*zapLogger).Error
> 	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/github.com/go-logr/zapr/zapr.go:128
> sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
> 	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:258
> sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
> 	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232
> sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
> 	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211
> k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
> 	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152
> k8s.io/apimachinery/pkg/util/wait.JitterUntil
> 	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:153
> k8s.io/apimachinery/pkg/util/wait.Until
> 	/data/dev/go_path/src/github.com/kiegroup/kie-cloud-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
> ```
> 
> Probably this step needs to be skipped on ocp version < 4.

@spolti this should be resolved

tchughesiv:
/assign @bmozaffa 

@spolti thoughts on whether we want to add this check?
https://github.com/kiegroup/kie-cloud-operator/blob/2fdcae9c54cc1d89c43b47adabc41a3c34561091/pkg/controller/initializer.go#L31-L33

And, if so, should 3.11 be designated as such?
https://github.com/kiegroup/kie-cloud-operator/blob/2fdcae9c54cc1d89c43b47adabc41a3c34561091/pkg/controller/kieapp/constants/constants.go#L22-L23

tchughesiv:
@bmozaffa hold off on your review... just remembered we wanted to use `golang.org/x/mod/semver` instead. Need to make those adjustments first.

tchughesiv:
/assign @bmozaffa 
Operator has been migrated to `golang.org/x/mod/semver`. It's worth noting though that our csv generation code continues to use `github.com/blang/semver`. This is because OLM uses that semver.

PR #426, opened 2020-05-04
fix jms and webhook password generation

tchughesiv:
With the new `Status.Applied` method, any passwords which we generate for the user must be carried forward between reconciles. Otherwise, an infinite loop occurs and reconciliation of objects never finishes.

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -404,26 +404,14 @@ func getServersConfig(cr *api.KieApp) ([]api.ServerTemplate, error) {
 	}
 	product := GetProduct(cr.Status.Applied.Environment)
 	usedNames := map[string]bool{}
-	unsetNames := 0
-	cr.Status.Applied.Objects.Servers = serverSortBlanks(cr.Status.Applied.Objects.Servers)
 	serverSlice := cr.Status.Applied.Objects.Servers
 	for index := range serverSlice {
 		serverSet := &serverSlice[index]
 		if serverSet.Deployments == nil {
 			serverSet.Deployments = Pint(constants.DefaultKieDeployments)
 		}
-		if serverSet.Name == "" {
-			for i := 0; i < len(serverSlice); i++ {
-				serverSetName := getKieSetName(cr.Status.Applied.CommonConfig.ApplicationName, serverSet.Name, unsetNames)
-				if !usedNames[serverSetName] {
-					serverSet.Name = serverSetName
-					break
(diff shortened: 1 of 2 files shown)


pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -911,7 +934,7 @@ func TestBuildConfiguration(t *testing.T) {
 	assert.Equal(t, "https://maven.mirror.com/", server.BuildConfigs[0].Spec.Strategy.SourceStrategy.Env[1].Value)
 	assert.Equal(t, "dir", server.BuildConfigs[0].Spec.Strategy.SourceStrategy.Env[2].Value)
 	assert.Equal(t, "s3cr3t", server.BuildConfigs[0].Spec.Triggers[0].GitHubWebHook.Secret)
-	assert.NotEmpty(t, server.BuildConfigs[0].Spec.Triggers[1].GenericWebHook.Secret)
+	assert.Empty(t, server.BuildConfigs[0].Spec.Triggers[1].GenericWebHook.Secret)
```
tchughesiv, line 937: why should this be `NotEmpty` by default?

tchughesiv: @ruromero @bmozaffa thoughts?

ruromero: IIRC it is generated by default to allow external apps to use this secret when sending webhook events. The buildConfig is the owner of the secret so it makes sense to have always a generated one by default

ruromero: But this should apply to the defined Webhook. In this case as the GitHubWebhook is defined I agree the secret for the GenericWebhook should not be generated.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -404,26 +404,15 @@ func getServersConfig(cr *api.KieApp) ([]api.ServerTemplate, error) {
...
-		if serverSet.Name == "" {
-			for i := 0; i < len(serverSlice); i++ {
-				serverSetName := getKieSetName(cr.Status.Applied.CommonConfig.ApplicationName, serverSet.Name, unsetNames)
-				if !usedNames[serverSetName] {
-					serverSet.Name = serverSetName
-					break
-				}
-				unsetNames++
-			}
-		}
```
tchughesiv, line 424: @ruromero one more change since your review...  I've decided to move this code to SetDefaults() and remove from here as redundant.

PR #427, opened 2020-05-06
fix logic for webhook secrets

tchughesiv:
a continuation of https://github.com/kiegroup/kie-cloud-operator/pull/426
newly discovered bugs created from the new `status.applied` section


tchughesiv:
@ruromero i've moved the of the webhook secret generation logic to a more central location will continue to create webhook secrets by default for both webhook types... just as we've always done in prior product versions.

PR #434, opened 2020-05-19
[RHPAM-2746 ] Set default JVM Max memory ratio to 80% in BA Operator

desmax74:
https://issues.redhat.com/browse/RHPAM-2746
@tchughesiv @spolti @jakubschwan

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -944,16 +944,29 @@ func SetDefaults(cr *api.KieApp) {
 		specApply.Objects.Servers = []api.KieServerSet{{Deployments: Pint(constants.DefaultKieDeployments)}}
 	}
 	setKieSetNames(specApply)
+	setJvmDefault(specApply.Objects.Console.Jvm)
 	for index := range specApply.Objects.Servers {
 		addWebhookTypes(specApply.Objects.Servers[index].Build)
 		for _, statusServer := range cr.Status.Applied.Objects.Servers {
 			retainAppliedPwds(&specApply.Objects.Servers[index], statusServer)
 		}
 		addWebhookPwds(specApply.Objects.Servers[index].Build)
+		setJvmDefault(specApply.Objects.Servers[index].Jvm)
 	}
 	isTrialEnv := strings.HasSuffix(string(specApply.Environment), constants.TrialEnvSuffix)
 	setPasswords(specApply, isTrialEnv)
 	cr.Status.Applied = *specApply
+}
+
+func setJvmDefault(jvm *api.JvmObject) {
+	if jvm != nil {
+		if jvm.JavaMaxMemRatio == nil {
+			jvm.JavaMaxMemRatio = Pint32(80)
+		}
+		if jvm.JavaInitialMemRatio == nil {
+			jvm.JavaInitialMemRatio = Pint32(25)
+		}
+	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -950,12 +950,26 @@ func SetDefaults(cr *api.KieApp) {
 			retainAppliedPwds(&specApply.Objects.Servers[index], statusServer)
 		}
 		addWebhookPwds(specApply.Objects.Servers[index].Build)
+		setJvmDefault(specApply.Objects.Console.Jvm)
```
tchughesiv, line 953: since this doesn't affect the `specApply.Objects.Servers` object, it can be moved it out of the for loop it's currently in... maybe up to line 946?

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -4381,3 +4381,70 @@ func TestMergeDBDeployment(t *testing.T) {
...
+		Value: "test",
+	}
+	cr := &api.KieApp{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: api.KieAppSpec{
+			Environment: api.RhdmTrial,
+			CommonConfig: api.CommonConfig{
+				ApplicationName: "trial",
```
tchughesiv, line 4402: this commonconfig section is probably unecessary for a jvm specific test.

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -4381,3 +4381,70 @@ func TestMergeDBDeployment(t *testing.T) {
...
+			},
+			Objects: api.KieAppObjects{
+				Console: api.ConsoleObject{
+					KieAppObject: api.KieAppObject{
+						Env: []corev1.EnvVar{
+							envReplace,
+							envAddition,
+						},
+					},
```
tchughesiv, line 4411: this kieappobject section is probably unecessary for a jvm specific test.

pkg/controller/kieapp/defaults/defaults.go
desmax74: Ok

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: Ok

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: Ok

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -944,6 +944,7 @@ func SetDefaults(cr *api.KieApp) {
 		specApply.Objects.Servers = []api.KieServerSet{{Deployments: Pint(constants.DefaultKieDeployments)}}
 	}
 	setKieSetNames(specApply)
+	setJvmDefault(specApply.Objects.Console.Jvm)
```
tchughesiv, line 947: should we also add one for kieserversets since they also have a jvm pointer like console does?
https://github.com/kiegroup/kie-cloud-operator/blob/2253fc607392f8a7f135dde1539832bc22643903/pkg/apis/app/v2/kieapp_types.go#L127

if so, it should likely be added to the for loop below.

desmax74: Ok, I'll add with test

PR #443, opened 2020-06-16
[RHPAM-2752] Add default CPU limit for images to guarantee product performance

desmax74:
@jakubschwan @spolti @tchughesiv

### DIFF
--- a/deploy/ui/form.json
+++ b/deploy/ui/form.json
@@ -558,8 +558,7 @@
                   "label": "CPU request",
                   "type": "text",
                   "jsonPath": "$.spec.objects.console.resources.requests.cpu",
-                  "description": "Minimum processing power requested by the container",
-                  "default": "500m"
+                  "description": "Minimum processing power requested by the container, default is 1000m"
                 },
                 {
                   "label": "Memory limit",
@@ -572,8 +571,7 @@
                   "label": "CPU limit",
                   "type": "text",
                   "jsonPath": "$.spec.objects.console.resources.limits.cpu",
-                  "description": "Maximum processing power granted to the container",
-                  "default": "500m"
+                  "description": "Maximum processing power granted to the container, default is 2000m"
                 }
               ]
             },
@@ -976,8 +974,7 @@
                       "label": "CPU request",
                       "type": "text",
                       "jsonPath": "$.spec.objects.servers[*].resources.requests.cpu",
-                      "default": "500m",
-                      "description": "Minimum processing power requested by the container"
+                      "description": "Minimum processing power requested by the container, default is 500m"
                     },
                     {
                       "label": "Memory limit",
@@ -990,8 +987,7 @@
                       "label": "CPU limit",
                       "type": "text",
                       "jsonPath": "$.spec.objects.servers[*].resources.limits.cpu",
-                      "default": "500m",
-                      "description": "Maximum processing power granted to the container"
+                      "description": "Maximum processing power granted to the container, default is 1000m"
                     }
                   ]
                 },
@@ -1597,8 +1593,7 @@
                       "label": "CPU request",
                       "type": "text",
                       "jsonPath": "$.spec.objects.smartRouter.resources.requests.cpu",
-                      "default": "500m",
-                      "description": "Minimum processing power requested by the container"
+                      "description": "Minimum processing power requested by the container, default is 250m"
                     },
                     {
                       "label": "Memory limit",
@@ -1611,8 +1606,7 @@
                       "label": "CPU limit",
                       "type": "text",
                       "jsonPath": "$.spec.objects.smartRouter.resources.limits.cpu",
-                      "default": "500m",
-                      "description": "Maximum processing power granted to the container"
+                      "description": "Maximum processing power granted to the container, default is 500m"
                     }
                   ]
                 }
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -17,12 +17,15 @@ import (
 	"github.com/gobuffalo/packr/v2"
 	"github.com/imdario/mergo"
 	api "github.com/kiegroup/kie-cloud-operator/pkg/apis/app/v2"
+	v2 "github.com/kiegroup/kie-cloud-operator/pkg/apis/app/v2"
 	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp/constants"
 	"github.com/kiegroup/kie-cloud-operator/pkg/controller/kieapp/shared"
 	"github.com/kiegroup/kie-cloud-operator/version"
 	"golang.org/x/mod/semver"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -966,9 +969,17 @@ func SetDefaults(cr *api.KieApp) {
 		}
 		addWebhookPwds(specApply.Objects.Servers[index].Build)
 		setJvmDefault(specApply.Objects.Servers[index].Jvm)
+		setResourcesDefault(&specApply.Objects.Servers[index].KieAppObject, constants.ServersCPULimit, constants.ServersCPURequests)
 	}
 	isTrialEnv := strings.HasSuffix(string(specApply.Environment), constants.TrialEnvSuffix)
 	setPasswords(specApply, isTrialEnv)
+
+	setResourcesDefault(&specApply.Objects.Console.KieAppObject, constants.ConsoleCPULimit, constants.ConsoleCPURequests)
+
+	if specApply.Objects.SmartRouter != nil {
+		setResourcesDefault(&specApply.Objects.SmartRouter.KieAppObject, constants.SmartRouterCPULimit, constants.SmartRouterCPURequests)
+	}
+
 	cr.Status.Applied = *specApply
 }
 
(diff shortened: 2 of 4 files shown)


deploy/crds/kieapp.crd.yaml
```diff
@@ -254,6 +254,19 @@ spec:
                             gcContainerOptions:
                               type: string
                               description: Specify Java GC to use. The value of this variable should contain the necessary JRE command-line options to specify the required GC, which will override the default of '-XX:+UseParallelOldGC'. e.g. '-XX:+UseG1GC'
+                        cpuLimit:
```
jakubschwan, line 257: I don't think that this configuration is correct.
1) We are already able to set limit for CPU (and for Memory) via resources in objects [e.g. for BC here](https://github.com/kiegroup/kie-cloud-operator/pull/443/files#diff-666b97114f2f43894579e681aaa0fcc9R154-R160). If we are OK with adding this new cpuLimit then we should cover in test cross configuration and set what has the bigger priority for configuration - I guess that this should be more like a constant holder.
2) And this is more Important I think, CPU Limit should not be set for Objects - here for console and later is set again for Kie Server. Here in console setting it does not make sense to configure CPU limits for Kie Server and DataGrid. I am also worried that this can bring some misconfiguration later. This CPU Limits should be set on some higher lever of configuration e.g. in Spec

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -4461,6 +4461,60 @@ func createJvmTestObjectWithoutJavaMaxMemRatio() *api.JvmObject {
 	return &jvmObject
 }
 
+func TestCPUDefaultLimit(t *testing.T) {
```
jakubschwan, line 4464: Can you please also add these test scenario:
1) Test case when are set CPU Limits (not default values)
2) Set CPU Limits in different form - e.g. instead of ms use cores (2000m = 2)
3) Test case when are set at the same time CPU Limits and Object's Resources limits with different values
4) Test case when is set Object's Resources request higher then CPU Limit (no Resource limit)

deploy/crds/kieapp.crd.yaml
desmax74: Here we set the cpu limit on BC, KieServer and DataGrid https://github.com/jboss-container-images/rhpam-7-openshift-image/blob/master/templates/rhpam78-authoring-ha.yaml, if we set in Spec could we use different values for different pods ?

deploy/ui/form.json
```diff
@@ -976,7 +976,7 @@
                       "label": "CPU request",
                       "type": "text",
                       "jsonPath": "$.spec.objects.servers[*].resources.requests.cpu",
-                      "default": "500m",
+                      "default": "1000m",
```
spolti, line 979: for kieserver I would stick minimum as 500m and 1000m as maximum.

deploy/ui/form.json
```diff
@@ -1597,7 +1597,7 @@
                       "label": "CPU request",
                       "type": "text",
                       "jsonPath": "$.spec.objects.smartRouter.resources.requests.cpu",
-                      "default": "500m",
+                      "default": "1000m",
```
spolti, line 1600: smartrouter does not need such value, maybe 250m for minimum and 500 for maximum.

deploy/crs/v2/kieapp_rhdm_authoring-ha.yaml
```diff
@@ -8,3 +8,16 @@ metadata:
     consoleDesc: Deploys a DM Authoring HA environment
 spec:
   environment: rhdm-authoring-ha
+  objects:
```
spolti, line 11: why instead doing it on every env.yaml file you don't do this at code lvl? it would save you a few lines and is better imho.
like here for example: https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/controller/kieapp/constants/constants.go

Or here: https://github.com/kiegroup/kie-cloud-operator/blob/81534c293f1df47d19113f5e6e749ccc080d2b7a/config/7.8.0/common.yaml#L43

@tchughesiv what do you suggest in this case?

tchughesiv: Agree... @desmax74 and I have discussed this same thing.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -966,9 +968,12 @@ func SetDefaults(cr *api.KieApp) {
 		}
 		addWebhookPwds(specApply.Objects.Servers[index].Build)
 		setJvmDefault(specApply.Objects.Servers[index].Jvm)
+		setResourcesDefaultServers(specApply.Objects.Servers[index].Resources)
```
jakubschwan, line 971:
```suggestion
		setResourcesDefaultServer(specApply.Objects.Servers[index].Resources)
```

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -983,6 +988,59 @@ func setJvmDefault(jvm *api.JvmObject) {
 	}
 }
 
+func setResourcesDefaultServers(resources *corev1.ResourceRequirements) {
```
jakubschwan, line 991:
```suggestion
func setResourcesDefaultServer(resources *corev1.ResourceRequirements) {
```

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -966,9 +968,12 @@ func SetDefaults(cr *api.KieApp) {
 		}
 		addWebhookPwds(specApply.Objects.Servers[index].Build)
 		setJvmDefault(specApply.Objects.Servers[index].Jvm)
+		setResourcesDefaultServers(specApply.Objects.Servers[index].Resources)
 	}
 	isTrialEnv := strings.HasSuffix(string(specApply.Environment), constants.TrialEnvSuffix)
 	setPasswords(specApply, isTrialEnv)
+	setResourcesDefaultConsole(&specApply.Objects.Console)
+	setResourcesDefaultSmartRouter(specApply.Objects.SmartRouter)
```
jakubschwan, line 976: Wouldn't be better to keep the same format of settings the Resources defaults for all object? We keep the same format for setJvmDefault. Also there could be only one function for settting up the default Resoruces and we could give to it also the constants that's need to be set.
```suggestion
	setResourcesDefaultConsole(&specApply.Objects.Console.Resources)
	setResourcesDefaultSmartRouter(specApply.Objects.SmartRouter.Resources)
```

pkg/controller/kieapp/test/crd_validation_test.go
```diff
@@ -184,9 +184,9 @@ func snippets(t *testing.T, file, yamlStr string) string {
 
 func TestJvmCrd(t *testing.T) {
 	crdYaml, _ := packr.New("deploy/crds", "../../../../deploy/crds").FindString("kieapp." + api.SchemeGroupVersion.Version + ".crd.yaml")
-	crdJson, _ := yaml.YAMLToJSON([]byte(crdYaml))
+	crdJSON, _ := yaml.YAMLToJSON([]byte(crdYaml))
```
jakubschwan, line 187: this test changes for JVM test are not related to default CPU settings, are they really needed?

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/controller/kieapp/test/crd_validation_test.go
desmax74: No, I'll remove this fix

config/7.8.0/envs/rhdm-authoring-ha.yaml
```diff
@@ -116,10 +116,8 @@ others:
                     timeoutSeconds: 10
                   resources:
                     limits:
-                      cpu: "1000m"
```
tchughesiv, line 119: datagrid, i believe, is one object that isn't yet configurable with resource limits. so, we likely need to keep these settings for the datagrid statefulsets in rhdm/pam authoring-ha envs.

config/7.8.0/envs/rhdm-authoring-ha.yaml
```diff
@@ -116,10 +116,8 @@ others:
...
                       memory: "2Gi"
                     requests:
-                      cpu: "1000m"
```
tchughesiv, line 122: and here...

config/7.8.0/envs/rhpam-authoring-ha.yaml
```diff
@@ -116,10 +116,8 @@ others:
                     timeoutSeconds: 10
                   resources:
                     limits:
-                      cpu: "1000m"
```
tchughesiv, line 119: and here...

config/7.8.0/envs/rhpam-authoring-ha.yaml
```diff
@@ -116,10 +116,8 @@ others:
...
                       memory: "2Gi"
                     requests:
-                      cpu: "1000m"
```
tchughesiv, line 122: and here...

deploy/ui/form.json
```diff
@@ -558,8 +558,7 @@
                   "label": "CPU request",
                   "type": "text",
                   "jsonPath": "$.spec.objects.console.resources.requests.cpu",
-                  "description": "Minimum processing power requested by the container",
-                  "default": "500m"
+                  "description": "Minimum processing power requested by the container"
```
spolti, line 561: Do you mind to improve the description here and for others below telling what are the default value for each case if no value is provided?

config/7.8.0/envs/rhdm-authoring-ha.yaml
jakubschwan: This configuration is set to the statefulsets (and set for all pods). As DataGrid is RH MW image it share the same cct_module scripts and the configuration applied here should be also handle by DataGrid image (in a same way as we are configuring e.g. Kie Server image). Am I right @spolti ? 
@desmax74 I think you can easily check if limits are applied to pods, when you deploy the app and take a look into DataGrid pod log, there should be -Xmx=1024 (As we set there memory limit to 2Gi and defautlt JVM mem ratio is set to 50%)

deploy/ui/form.json
desmax74: Done

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -983,6 +994,72 @@ func setJvmDefault(jvm *api.JvmObject) {
...
+
+		if server.Resources.Requests == nil {
+			server.Resources.Requests = createCPURequests(constants.ServersCPURequests)
+		}
+	} else {
+		server.Resources = createResourceWithLimitsAndRequests(constants.ServersCPULimit, constants.ServersCPURequests)
+	}
+}
+
+func setResourcesDefaultConsole(console *v2.ConsoleObject) {
```
tchughesiv, line 1012: instead of using a separate method for each object type... maybe instead we can have 1 method that all of them can use? Consider that KieAppObject is what actually houses the Resources struct and that KieAppObject exists in console, servers, and smartrouter. What if this method instead look like this? 
```
func setResourcesDefault(console *v2.KieAppObject, limits, requests string) {
```
Could console, servers and smartrouter all potentially use that same method?

PR #445, opened 2020-06-16
[RHPAM-2996] add ssh secret option to console object for githooks

tchughesiv:
mounts a secret in BC at `/home/jboss/.ssh` with the following required files -
 - id_rsa
 - known_hosts

### DIFF
--- a/config/7.8.0/common.yaml
+++ b/config/7.8.0/common.yaml
@@ -268,6 +268,10 @@ console:
                     mountPath: "[[.Console.GitHooks.MountPath]]"
                     readOnly: true
                   #[[end]]
+                  #[[if ne .Console.GitHooks.SSHSecret ""]]
+                  - name: "[[.Constants.GitHooksSSHSecret]]"
+                    mountPath: "/home/jboss/.ssh"
+                  #[[end]]
             volumes:
               - name: "[[.ApplicationName]]-[[.Console.Name]]-[[.Constants.KeystoreVolumeSuffix]]"
                 secret:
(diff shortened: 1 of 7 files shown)


config/7.8.0/common.yaml
```diff
@@ -268,6 +268,11 @@ console:
                     mountPath: "[[.Console.GitHooks.MountPath]]"
                     readOnly: true
                   #[[end]]
+                  #[[if .Console.GitHooks.SSHSecret]]
+                  - name: "[[.Constants.GitHooksSSHSecret]]"
+                    mountPath: "/home/jboss/.ssh"
+                    readOnly: true
```
spolti, line 274: my only concern about it is the permissions,
for directory it should be 700 and files 644.

tchughesiv: I've fixed the perms issue... but i'm also concerned about overwriting the user's .ssh dir. Is there no way the product can look in another dir for ssh keys so we can leave the ~/.ssh dir alone?
@mramendi ?

mramendi: @tchughesiv I have checked the 7.7.0 production BC image during out call. A /home/jboss/.ssh directory does not exist there, so there is nothing to overwrite.

mramendi: For safety, perhaps the directory itself should not be read-only? In case the product writes there at runtime?

tchughesiv:
@spolti @mramendi can one of you pass a `/lgtm`?

deploy/crds/kieapp.crd.yaml
```diff
@@ -182,6 +182,9 @@ spec:
                           type: object
                           description: GitHooks configuration object
                           properties:
+                            sshSecret:
+                              type: string
+                              description: Secret to use for ssh key(s) mount.
```
mramendi, line 187: The (s) is incorrect as we only use one key. Suggestion:
Secret to use for ssh key and known hosts file

deploy/ui/form.json
```diff
@@ -516,6 +516,13 @@
                       "description": "Reference object Name for the GitHooks."
                     }
                   ]
+                },
+                {
+                  "label": "SSH Secret",
+                  "type": "text",
+                  "required": false,
+                  "jsonPath": "$.spec.objects.console.gitHooks.sshSecret",
+                  "description": "Secret to use for ssh key(s) mount."
```
mramendi, line 525: The (s) is incorrect as we only use one key. Suggestion:
Secret to use for ssh key and known hosts file

PR #460, opened 2020-07-06
[RHPAM 2979] Add option for user set custom context image on Operator and UI

desmax74:
https://issues.redhat.com/browse/RHPAM-2979

@tchughesiv @jakubschwan


tchughesiv:
/hold
I believe this capability already exists - 
https://github.com/kiegroup/kie-cloud-operator/blob/83f8c629b52f363dec13843986537eb239807042/deploy/crds/kieapp.crd.yaml#L636-L645

deploy/crds/kieapp.crd.yaml
```diff
@@ -637,6 +643,9 @@ spec:
                       type: object
                       description: Configuration of the RHPAM smart router
                       properties:
+                        context:
```
tchughesiv, line 646: maybe `imageContext` is better here? to give user a better understanding of its purpose?

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -4640,3 +4641,206 @@ func TestSimplifiedMonitoringSwitch(t *testing.T) {
...
+		KieAppObject: api.KieAppObject{
+			ImageContext: "rhpam-42",
+		},
+		Protocol:         "",
+		UseExternalRoute: false,
+	}
+	return &smartRouter
+}
+
+func TestConsoleDefault(t *testing.T) {
```
tchughesiv, line 4707: might want to give these new default tests a slightly more descriptive name? `TestConsoleDefaultImage` maybe?

desmax74: ok

tchughesiv:
We’ll hold this until master is on 7.9 version.

tchughesiv:
@desmax74 since your cpu PR merged, there appears to be a conflict that needs resolving.

deploy/crds/kieapp.crd.yaml
```diff
@@ -117,6 +117,9 @@ spec:
                       type: object
                       description: Configuration of the RHPAM workbench
                       properties:
+                        imageContext:
+                          type: string
+                          description: The context to use for console.
```
tchughesiv, line 122: can we make this "The **image** context to use for console."?

deploy/crds/kieapp.crd.yaml
```diff
@@ -271,6 +274,9 @@ spec:
                           id:
                             type: string
                             description: Server ID
+                          imageContext:
+                            type: string
+                            description: The context to use for server.
```
tchughesiv, line 279: same

deploy/crds/kieapp.crd.yaml
```diff
@@ -637,6 +643,9 @@ spec:
                       type: object
                       description: Configuration of the RHPAM smart router
                       properties:
+                        imageContext:
+                          type: string
+                          description: The context to use for smart router.
```
tchughesiv, line 648: same

deploy/crds/kieapp.crd.yaml
```diff
@@ -694,6 +703,9 @@ spec:
                       type: object
                       description: Configuration of the RHPAM Process Instance Migration
                       properties:
+                        imageContext:
+                          type: string
+                          description: The imageContext to use for Process Instance Migration.
```
tchughesiv, line 708: same

swati-kale:
@desmax74 hi, I want to push the UI changes in this branch, but you might need to give me permission or I will submit separate PR for UI.  Thanks.

jakubschwan:
@spolti can you please take a look on this PR as well and provide feedback, thanks :)

spolti:
I agree with you @jakubschwan 
@tchughesiv @desmax74 wdyt?

desmax74:
@jakubschwan @spolti @tchughesiv About the position, the Image, ImageTag and Image Context are related and is correct to stay close, agree about examples near description like "rhpam-7/rhpam-kieserver-rhel8:7.9.0 where rhpam-7 is the context, rhpam-kieserver-rhel8 is the image and 7.9.0 is the image tag, you could use your providing custom values", agree to describe which are the default values, I don't have opinion about button to enable the customization, since the other two parts (image and image tag) are simple input box.

swati-kale:
> @jakubschwan @spolti @tchughesiv About the position, the Image, ImageTag and Image Context are related and is correct to stay close, agree about examples near description like "rhpam-7/rhpam-kieserver-rhel8:7.9.0 where rhpam-7 is the context, rhpam-kieserver-rhel8 is the image and 7.9.0 is the image tag, you could use your providing custom values", agree to describe which are the default values, I don't have opinion about button to enable the customization, since the other two parts (image and image tag) are simple input box.

swati-kale:
@jakubschwan 
1)"Add/Delete buttons " are for object arrays, But we can group these fields together in fieldgroup and provide a checkbox to the user to choose the configuration. Let me know.

2) For Kieservers:: Do you want these 3 fields inside group "Kieserver Image" and after the fields "NameSpace"?

jakubschwan:
### 2 ways of configuration of image for Kie Server
Right now for Kie Server we have 2 ways how to set up the Image that should be used for the Deployment Config. I will try to summarize both ways and their conflicts and issues, that I noticed.
Both options are used to generate imageUrl, that is applied to the Kie Server. - https://github.com/desmax74/kie-cloud-operator/blob/rhpam-2979/config/7.9.0/common.yaml#L610 
Both configuration options are checked in one method - https://github.com/desmax74/kie-cloud-operator/blob/rhpam-2979/pkg/controller/kieapp/defaults/defaults.go#L687 
When we used both configuration options, then was used configuration set by “server.from“.

#### Setting image by `server.from` object

- This object has two options for Image deploying - ImageStreamTag or DockerImage
- For both we need to specify name. Namespace is optional.
- Name should be in format “<imageName>:<imageTag>”
- If image is not found, then Operator tries to create and pull it from default RH registry (registry.redhat.io/rhpam-7/) 
- If an image was not found, ImageStream is not valid.
- DeploymentConfig will not start (expected)
- With valid image tag, DeploymentConfig works
- Tried only ImageStreamTag option

#### Setting image by `sever.image` `server.imageTag` `server.imageContext`
- Image context is added as new option
- Tried to configure it for the same image stream as for the `server.form`
- Image stream was not founded - it tried to look in openshift namespace
  - THIS IS A BUG
  - Operator tried to look up for image “openshift/<imageContext>/<imageName>:<imageTag>”
  - After that it tried to create image (this time in correct place) “<imageContext>/<imageName>:<imageTag>”
    - Creation failed as in name is not allowed char -  may not contain '/'
    - It was still tried to use default redhat registry - it tried to pull image “registry.redhat.io/rhpam-7/<imageContext>/<imageName>:<imageTag>”

deploy/ui/form.json
```diff
@@ -1547,6 +1589,27 @@
                   "type": "text",
                   "description": "The secret name within the Java keystore used for https encryption. Generated if empty."
                 },
+                {
+                  "label": "Image Context",
+                  "required": false,
+                  "jsonPath": "$.spec.objects.smartRouter.imageContext e.g. rhpam-7, this param is optional for custom image.",
+                  "description": "The image context to use for smart router.",
```
spolti, line 1596: what context stands for, namespace or registry? if so we should update it.

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -229,6 +229,8 @@ type KieAppObject struct {
 	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
 	// Keystore secret name
 	KeystoreSecret string `json:"keystoreSecret,omitempty"`
+	// The image context to use
+	ImageContext string `json:"imageContext,omitempty"`
 	// The image to use
 	Image string `json:"image,omitempty"`
 	// The image tag to use
```
jakubschwan, line 236: Could you please add more specific description for these parmas? Like that this parameter is optional, in what cases is needed to be set, example value.
I think we can copy description from UI here, then it would be visible also in crd, as crd is generated from api :)

desmax74: Done

deploy/ui/form.json
desmax74: Was a part of description in the wrong place, updated

PR #468, opened 2020-07-13
[RHPAM-3055] Use CR name in the console links

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>

Fix https://issues.redhat.com/browse/RHPAM-3055

### DIFF
--- a/pkg/controller/kieapp/kieapp_controller_test.go
+++ b/pkg/controller/kieapp/kieapp_controller_test.go
@@ -787,7 +787,7 @@ func TestConsoleLinkCreation(t *testing.T) {
 
 	consoleLink := &consolev1.ConsoleLink{}
 	reconciler.Service.Get(context.TODO(), getNamespacedName("", getConsoleLinkName(cr)), consoleLink)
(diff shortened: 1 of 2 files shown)


pkg/controller/kieapp/kieapp_controller_test.go
```diff
@@ -787,7 +787,7 @@ func TestConsoleLinkCreation(t *testing.T) {
 
 	consoleLink := &consolev1.ConsoleLink{}
 	reconciler.Service.Get(context.TODO(), getNamespacedName("", getConsoleLinkName(cr)), consoleLink)
-	assert.Equal(t, "Business Central", consoleLink.Spec.Text)
+	assert.Equal(t, "cr: Business Central", consoleLink.Spec.Text)
```
tchughesiv, line 790: might this be better? to test new func?
```golang
assert.Equal(t, getConsoleLinkFriendlyName(cr), consoleLink.Spec.Text)
```

ruromero: I don't think so because the function can be wrong and return always "foo" and the test would still pass

PR #469, opened 2020-07-14
[RHPAM-2746] Set default JVM Max memory ratio to 80% in BA Operator

desmax74:
https://issues.redhat.com/browse/RHPAM-2746

@tchughesiv @jakubschwan


desmax74:
@tchughesiv  why is checking duplication on code not in the PR ?

tchughesiv:
@desmax74 it checks dup against all modified **files** i believe. maybe you duplicated some existing code?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1015,6 +1017,18 @@ func setJvmDefault(jvm *api.JvmObject) {
 	}
 }
 
+func checkJvmOnConsole(console *api.ConsoleObject) {
+	if console.Jvm == nil {
+		console.Jvm = &api.JvmObject{}
+	}
+}
+
+func checkJvmOnServer(server *api.KieServerSet) {
```
tchughesiv, line 1026: instead of creating these new functions... could we just add this logic here?
https://github.com/kiegroup/kie-cloud-operator/blob/5bae06a7c007fac6ab446c129a4346cdd84b55cd/pkg/controller/kieapp/defaults/defaults.go#L1009-L1018

desmax74:
@tchughesiv sonar indicate duplication on kieapp_controller_test.go and isn't code in this PR (defaults.go and defaults_test.go)

pkg/controller/kieapp/defaults/defaults.go
desmax74: If the jvm is nil we don't know to which object add the empty jvm struct, the setJvmDefault receive just a pointer

tchughesiv:
@desmax74 since your cpu PR merged, there appears to be a conflict that needs resolving.

PR #492, opened 2020-08-17
[RHPAM-3476] use openshift CA for truststore generation

tchughesiv:
Use openshift's ca-bundle as the truststore for PAM java components. Enable w/ `spec.truststore.openshiftCaBundle`.
This will mount the openshift ca-bundle as a truststore secret to the following pam objects -
- console
- servers
- dashbuilder
- smartrouter

https://issues.redhat.com/browse/RHPAM-3476


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -29,7 +29,14 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-var log = logs.GetLogger("kieapp.defaults")
+var (
+	log          = logs.GetLogger("kieapp.defaults")
+	caOptsAppend = []string{
```
spolti, line 34: in case the user needs to provide a new TrustStore to access internal services outside ocp, is there a way to append the user's provided TrustStore to this one which will be used to configure the JVM?

spolti (changes requested):
just a minor comment.

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: i don't believe so... would be an either or kind of thing

tchughesiv: that said, if they added their trusted CA certs to the openshift cluster

deploy/ui/form.json
```diff
@@ -55,6 +55,14 @@
           "jsonPath": "$.spec.useImageTags",
           "originalJsonPath": "$.spec.useImageTags"
         },
+        {
+          "label": "Use Openshift CA",
+          "type": "checkbox",
+          "required": false,
+          "description": "Set true to use the Openshift CA bundle as a truststore. Replaces java's cacert.",
```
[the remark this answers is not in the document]
tchughesiv: no, just generates a truststore. we can't create certs against a CA w/o knowing the root CA privateKey. that's still something they'd have to do themselves.

are you sure about the custom truststore not being used instead of cacerts?

pkg/controller/kieapp/defaults/defaults.go
spolti: ok, we might need to add a note about it on the docs just to make users aware and update it on future releases, wdyt?

deploy/ui/form.json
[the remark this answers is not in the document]
spolti: sry man, I always get confused about the usage of the SSL related properties. 
When the trustStore property is set, the JVM, instead of looking on the default location for JVM's default cacerts will look on the provided location. you're right.

wondering If would make sense to instead set the properties, merge the ocp's on the jvm's one, does it make sense to you? If so, we can add a new task for a future release.

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: sure... i mean, maybe we could add a way to add a user-provided truststore as well in the future? but it felt like this was a big step forward using openshift's ca bundle.

deploy/ui/form.json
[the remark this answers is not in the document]
tchughesiv: this wouldn't work, the operator won't have access to java's cacert as it's bundled within the image. the only way that becomes possible is either with some kind of init container and keytool... or with a buildconfig in openshift. 

i wouldn't recommend either path, and honestly, openshift's ca bundle appears to support more CAs w/ 130 certs... whereas java's `cacert` file only has 93 certs.

pkg/controller/kieapp/defaults/defaults.go
spolti: yep, let's track it on a future improvement.

deploy/ui/form.json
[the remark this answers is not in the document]
spolti: or we can edit the image itself to search for mounted truststore and append it, maybe. My main concern is when the user needs to provide their own cert but does not have enough permissions to edit the OCP's ca-bundle, but, as discussed below, it can be done later.

PR #493, opened 2020-08-17
[KIECLOUD-417] Use sdk to generate KieApp CRD

tchughesiv:


bmozaffa:
For review purposes, can we make it so in at least one commit, if not the entire PR, we avoid renaming the CRD and can compare the new generated version to the old one?

deploy/crds/kieapp.crd.yaml
```diff
@@ -10,1768 +10,3302 @@ spec:
...
-                            name:
-                              type: string
-                              description: Object name
-                            namespace:
-                              type: string
-                              description: Namespace where the object is located
                         build:
-                          type: object
-                          description: Configuration of build configs for immutable KIE servers
-                          oneOf:
```
tchughesiv, line 313: I think I've got everything replicated from our old CRD, except for a couple of these `oneOf` schema options. I've yet to find a validation marker we can use to generate this section. @ruromero how critical are these?

tchughesiv:
@bmozaffa should be easier to review now.

PR #499, opened 2020-08-20
[KIECLOUD-413] - Increase AMQ broker version to 7.7

spolti:


tools/csv-gen/csv-gen.go
```diff
@@ -332,6 +332,7 @@ func main() {
 		relatedImages = addRefRelatedImages(constants.Datagrid73ImageURL16, constants.Datagrid73Component, imageRef, relatedImages)
 		relatedImages = addRefRelatedImages(constants.Broker75ImageURL, constants.BrokerComponent, imageRef, relatedImages)
```
tchughesiv, line 333: will `Broker75` still be used by any of the supported product versions in 7.9? if not, please remove from csv/constants files and regenerate csv.

spolti: I don't think so, ok, will remove it.

spolti: updated.

PR #503, opened 2020-09-02
[KIECLOUD-422] display admin user & pwd in Operator Hub

tchughesiv:
Leverage `urn:alm:descriptor:com.tectonic.ui:password` specDescriptor.
https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#8-password

Screen Shot 2020-09-18 at 2 04 41 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,14 +30,15 @@ If pushing to another quay repository, replace _kiegroup_ with your username or 
 **Create your own index image**
 ```bash
 $ make bundle-dev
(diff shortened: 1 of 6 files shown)


bmozaffa:
How would you populate this based on different scenarios? I was thinking (more recently) that maybe a field called something like "Login instructions" would be more appropriate, and can hold different values depending on what we want to say.

PR #511, opened 2020-09-22
upgrade fix: retain version from applied status

tchughesiv:
this is required now that we switched to status subresource.

/assign @bmozaffa 

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -1002,11 +1002,14 @@ func SetDefaults(cr *api.KieApp) {
 			api.SchemeGroupVersion.Group: version.Version,
 		})
 	}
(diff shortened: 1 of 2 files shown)


bmozaffa:
My two questions are, why is this required now that we've moved to sub resource, and if we shouldn't also do this with the credentials?

tchughesiv:
> My two questions are, why is this required now that we've moved to sub resource, and if we shouldn't also do this with the credentials?

1) i guess this technically has less to do with the move to subresource... and more to do with the switch to `status.applied`. This didn't affect us during upgrade in 7.8 because folks on 7.7 or prior still had version set in `spec`. Now testing upgrades to 7.9 i'm hitting issues where it upgrades product to 7.9 regardless of whether i want it to or not.
2) we already do this with credentials

PR #515, opened 2020-09-23
[KIECLOUD-407] - Operator rhpam-production-immutable fails due to Bus…

spolti:
…iness Monitoring PVC ReadWriteMany


pkg/apis/app/v2/kieapp_types.go
```diff
@@ -595,6 +595,7 @@ type TemplateConstants struct {
 // ConsoleTemplate contains all the variables used in the yaml templates
 type ConsoleTemplate struct {
 	OmitImageStream  bool           `json:"omitImageStream"`
+	Omit             bool           `json:"omit"`
```
tchughesiv, line 598: This shouldn't be necessary, since we already have an Omit option for pam components. For reference -
https://github.com/kiegroup/kie-cloud-operator/blob/5b00d9ab040857258aa17b10a679c57e4f09c1c9/pkg/apis/app/v2/kieapp_types.go#L538-L539
https://github.com/kiegroup/kie-cloud-operator/blob/5b00d9ab040857258aa17b10a679c57e4f09c1c9/pkg/controller/kieapp/defaults/defaults.go#L79-L81

config/7.9.0/envs/rhpam-production-immutable.yaml
```diff
@@ -1,3 +1,7 @@
+#[[if .Console.Omit]]
+console:
+  omit: true
+#[[end]]
```
tchughesiv, line 4: If we want the console omission to depend on number of replicas set, maybe adding this to the yaml templates doesn't make sense?

tchughesiv (changes requested):

config/7.9.0/envs/rhpam-production-immutable.yaml
tchughesiv: is this required for it to work?

spolti: like the yaml ig gt Console.Replicas?

tchughesiv: i actually meant the go code you are working on in this PR that checks for replicas.

spolti: updated it, wdyt?

tchughesiv: hmm... where'd the go code go? my thinking was get rid of this yaml addition and instead handle the omission logic in go.

spolti: if the plan is to remove these yaml, I do agree to keep the logic on the go side.
let me edit it.

spolti: Updated.

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -425,7 +425,39 @@ func TestRhpamProdImmutableEnvironment(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test",
+		},
+
+		Spec: api.KieAppSpec{
+			Environment: api.RhpamProductionImmutable,
+			Objects: api.KieAppObjects{
+				Console: api.ConsoleObject{
+					KieAppObject: api.KieAppObject{
+						Replicas: &replicas,
```
tchughesiv, line 447: can we add another test where you set replicas to 0 and test for Omit = true? Also, can you please test run this in a live cluster to ensure nothing with the other envs break?

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults_test.go
tchughesiv: hmm... wait, ignore that new unit test. it won't work and doesn't fit here. but def make sure envs work as expected.

spolti: I did run only the immutable on ocp, let me try the others one.

spolti:
Hey @tchughesiv please take a look when you have time.
updated it according our discussion on gchat. if no console is set for immutable rhpam env nothing related to it is created and ro rh dm immutable, nothing is created even if set on cr.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -217,6 +240,13 @@ func findCustomObjectByName(template api.CustomObject, objects []api.CustomObjec
 }
 
 func getEnvTemplate(cr *api.KieApp) (envTemplate api.EnvTemplate, err error) {
+
+	if !isImmutable(cr) && cr.Spec.Objects.Console == nil {
+		cr.Spec.Objects.Console = &api.ConsoleObject{
+			KieAppObject: api.KieAppObject{},
+		}
+	}
+
```
tchughesiv, line 249: Seems like this belongs in a different method? Maybe setDefaults?

tchughesiv, line 249: I think this makes more sense in the SetDefaults() method and should be applied to the `specApply` "defaulting" object.

We assign defaults to `status.applied` now instead of directly to `spec`.

tchughesiv (changes requested):

pkg/controller/kieapp/kieapp_controller.go
```diff
@@ -316,11 +316,14 @@ func setDeploymentStatus(instance *api.KieApp, resources []resource.KubernetesRe
 
 func (reconciler *Reconciler) verifyExternalReferences(cr *api.KieApp) error {
 	var err error
+
 	if cr.Status.Applied.Auth != nil && cr.Status.Applied.Auth.RoleMapper != nil {
 		err = reconciler.verifyExternalReference(cr.GetNamespace(), cr.Status.Applied.Auth.RoleMapper.From)
 	}
-	if err == nil && cr.Status.Applied.Objects.Console.GitHooks != nil {
-		err = reconciler.verifyExternalReference(cr.GetNamespace(), cr.Status.Applied.Objects.Console.GitHooks.From)
+	if cr.Spec.Objects.Console != nil {
```
tchughesiv, line 323: should be `cr.Status.Applied.Objects.Console`

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -80,10 +80,33 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+					_, found := shared.Find(constants.Envs2removeImmutableWithoutConsole, envvar.Name)
+					if !found {
+						newSlice = append(newSlice, envvar)
+					}
+				}
+				dc.Spec.Template.Spec.Containers[0].Env = newSlice
+			}
+		}
+	}
+
```
tchughesiv, line 109: This is something that might make more sense to handle in the go template configs. Instead of removing them after the fact, just never set them if the Console Object doesn't exist.

Something similar to this?
https://github.com/kiegroup/kie-cloud-operator/blob/e16f518f65bccd1f01ecf955fdf30b1ce87a4dc6/config/7.9.0/common.yaml#L200-L203

tchughesiv: maybe can do this?
#[[if not .Console.Omit]]

spolti: I avoid to touch the templates because you said that you was planning to remove some logic from templates.
I can move it to the template config if you think better.

pkg/controller/kieapp/defaults/defaults.go
spolti: I think I tired to add there but got some null pointer, will try again.

pkg/controller/kieapp/defaults/defaults.go
spolti: applied.

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: should assign to specApply

tchughesiv (changes requested):

pkg/controller/kieapp/kieapp_controller.go
spolti: applied.

pkg/controller/kieapp/constants/constants.go
```diff
@@ -21,6 +21,10 @@ var SupportedVersions = []string{CurrentVersion, PriorVersion1, PriorVersion2}
 // Ocp4Versions - OpenShift minor versions used for image curation
 var Ocp4Versions = []string{"4.5", "4.4", "4.3", "4.2", "4.1"}
 
+// Envs2removeImmutableWithoutConsole - Holds the environment variables that don't need to be set on Immutable
+// environments that does not contains Monitoring Console.
+// var Envs2removeImmutableWithoutConsole = []string{"WORKBENCH_SERVICE_NAME", "KIE_SERVER_CONTROLLER_PROTOCOL", "KIE_SERVER_CONTROLLER_SERVICE"}
+
```
tchughesiv, line 27: don't forget to remove this

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1002,9 +1023,17 @@ func SetDefaults(cr *api.KieApp) {
 			api.SchemeGroupVersion.Group: version.Version,
 		})
 	}
+
 	// retain certain items from status... e.g. version, usernames, passwords, etc
 	// everything else in status should be recreated with each reconcile.
 	specApply := cr.Spec.DeepCopy()
+
+	if !isImmutable(cr) && cr.Spec.Objects.Console == nil {
```
tchughesiv, line 1031: should be `specApply.Objects.Console`

tchughesiv (changes requested):

pkg/controller/kieapp/constants/constants.go
spolti: oops, right.

pkg/controller/kieapp/defaults/defaults.go
spolti: +1

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -1525,7 +1625,7 @@ func TestKieAppDefaults(t *testing.T) {
 
 	assert.False(t, cr.Spec.Upgrades.Enabled)
 	assert.Empty(t, cr.Spec.CommonConfig.ApplicationName)
-	assert.Nil(t, cr.Spec.Objects.Console.Replicas)
+	assert.Nil(t, cr.Spec.Objects.Console)
```
tchughesiv, line 1628: please check `cr.Status.Applied.Objects.Console` instead

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -80,10 +80,21 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 		env.SmartRouter.Omit = true
 	}
 
+	if cr.Spec.Environment == api.RhdmProductionImmutable || (isImmutable(cr) && cr.Status.Applied.Objects.Console == nil) {
```
tchughesiv, line 83: should be `cr.Status.Applied.Environment`

tchughesiv (changes requested):

tchughesiv (approved):

PR #523, opened 2020-10-01
[RHPAM-3297] Update metering labels

desmax74:
Update Metering labels

https://issues.redhat.com/browse/RHPAM-3297


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -131,33 +131,54 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 }
 
 func setProductLabels(cr *api.KieApp, env *api.Environment) {
-	setObjectLabels(cr, &env.Console, "business-central")
+	envConstants, _ := constants.EnvironmentConstants[cr.Status.Applied.Environment] // is this the only way to read the Image name/s ?
+	setObjectLabels(cr, &env.Console, "business-central", constants.BcImageName, getSubComponentTypeByImageName(constants.BcImageName))
```
tchughesiv, line 135: Not sure this is solution because image name can also have an rhdm prefix ... or maybe the user is leveraging a custom image with a different name. I think we need to be more dynamic in how we determine this.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -131,33 +131,54 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+			retValue = "infrastructure"
+	}
+	return retValue
+}
+
+
+func setObjectLabels(cr *api.KieApp, object *api.CustomObject, component string, subcomponent string, subcomponentType string) {
 	for index, obj := range object.DeploymentConfigs {
-		object.DeploymentConfigs[index].Spec.Template.Labels = setLabels(cr, obj.Spec.Template.Labels, component)
+		object.DeploymentConfigs[index].Spec.Template.Labels = setLabels(cr, obj.Spec.Template.Labels, component, subcomponent, subcomponentType)
```
tchughesiv, line 164: Would something like this be a more accurate approach?
```golang
		subcomponent, _, _ := GetImage(object.DeploymentConfigs[index].Spec.Template.Spec.Containers[0].Image)
		object.DeploymentConfigs[index].Spec.Template.Labels = setLabels(cr, obj.Spec.Template.Labels, component, subcomponent, getSubComponentTypeByImageName(subcomponent))
```

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
desmax74: We could consider "application" every image name different to SmartRouter and Controller, is it expected to use metering with custom images ?

pkg/controller/kieapp/defaults/defaults.go
desmax74: Sounds good, but how to retrieve the components who replace the single component, previously kie-server ?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -131,33 +131,51 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+	retValue := "application"
+	if imageName == constants.SmartRouterImageName || imageName == constants.ControllerImageName {
+		retValue = "infrastructure"
+	}
+	return retValue
 }
 
-func setObjectLabels(cr *api.KieApp, object *api.CustomObject, component string) {
+
+func setObjectLabels(cr *api.KieApp, object *api.CustomObject, component string, subcomponentType string) {
```
tchughesiv, line 155: no longer need to accept `subcomponentType` here

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
desmax74: correct, fixed

pkg/controller/kieapp/constants/constants.go
```diff
@@ -172,6 +174,12 @@ const (
 	ServersCPURequests     = "500m"
 	SmartRouterCPULimit    = "500m"
 	SmartRouterCPURequests = "250m"
+
+	//ImageNames for metering labels
+ 	BcImageName = RhpamPrefix + "-businesscentral-" + RhelVersion
+ 	ProcessMigrationImageName = RhpamPrefix + "-process-migration-" + RhelVersion
```
tchughesiv, line 180: Are the above constants still needed?

desmax74: No, removed

desmax74: No, removed

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -131,33 +131,48 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+	setObjectLabels(cr, &env.SmartRouter)
+
+	setObjectLabels(cr, &env.ProcessMigration)
+}
+
+func getSubComponentTypeByImageName(imageName string) string {
+	retValue := "application"
+	if imageName == constants.SmartRouterImageName || imageName == constants.ControllerImageName {
+		retValue = "infrastructure"
+	}
```
tchughesiv, line 147: what about rhdm images?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -131,33 +131,48 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
 
-func setLabels(cr *api.KieApp, labels map[string]string, component string) map[string]string {
+func setLabels(cr *api.KieApp, labels map[string]string, subcomponent string, subcomponentType string) map[string]string {
 	if labels == nil {
 		labels = map[string]string{}
 	}
 	labels[constants.LabelRHproductName] = constants.ProductName
 	labels[constants.LabelRHproductVersion] = cr.Status.Applied.Version
-	labels[constants.LabelRHcomponentName] = component
+	labels[constants.LabelRHcomponentName] = "PAM"
```
tchughesiv, line 171: what about DM?

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
desmax74: Added

pkg/controller/kieapp/defaults/defaults.go
desmax74: Is under the same umbrella label

jakubschwan (approved):

tchughesiv (approved):

PR #532, opened 2020-10-26
[RHDM-1419] - Immutable KieServer fails verifying kjar if configured …

spolti:
…maven uses self signed certificate


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -127,9 +127,25 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 		return api.Environment{}, err
 	}
 	setProductLabels(cr, &mergedEnv)
+	setCRBuildConfigEnvsToSpec(cr, &mergedEnv)
```
tchughesiv, line 130: any chance this can be moved into this method instead?
https://github.com/kiegroup/kie-cloud-operator/blob/d3cab494497a791b417306581623963cea7b2604/pkg/controller/kieapp/defaults/defaults.go#L618

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -127,9 +127,25 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
 
+func setCRBuildConfigEnvsToSpec(cr *api.KieApp, env *api.Environment) {
+	if len(env.Servers) > 0 {
+		for index := range cr.Spec.Objects.Servers {
+			for bcindex, srv := range env.Servers[index].BuildConfigs {
+				if srv.Spec.Strategy.SourceStrategy.Env != nil && cr.Spec.Objects.Servers[index].Build != nil {
+					for _, crenv := range cr.Spec.Objects.Servers[index].Build.Envs {
+						env.Servers[index].BuildConfigs[bcindex].Spec.Strategy.SourceStrategy.Env = append(
+							env.Servers[index].BuildConfigs[bcindex].Spec.Strategy.SourceStrategy.Env, crenv)
+					}
```
tchughesiv, line 142: we already have a method that i believe will do what you need ... 
https://github.com/kiegroup/kie-cloud-operator/blob/d3cab494497a791b417306581623963cea7b2604/pkg/controller/kieapp/defaults/defaults.go#L636
https://github.com/kiegroup/kie-cloud-operator/blob/d3cab494497a791b417306581623963cea7b2604/pkg/controller/kieapp/shared/utils.go#L145-L156

could it maybe go here?
https://github.com/kiegroup/kie-cloud-operator/blob/d3cab494497a791b417306581623963cea7b2604/pkg/controller/kieapp/defaults/defaults.go#L625-L628

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
spolti: The consolidateObjects will be called only during the reconcile, wouldn't better configure it before?

spolti: update the func to use EnvOverride.

tchughesiv: everything the operator does is ultimately only called during the reconcile. my main concern is to keep alike code together so we can find things more easily

spolti: I see, updated.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -625,10 +627,24 @@ func ConsolidateObjects(env api.Environment, cr *api.KieApp) api.Environment {
 	for index := range env.Servers {
 		serverSet, _ := GetServerSet(cr, index)
 		env.Servers[index] = ConstructObject(env.Servers[index], serverSet.KieAppObject)
+		setCRBuildConfigEnvsToSpec(cr, &env)
```
tchughesiv, line 630: since, at this point in code, we already have the server index... should we instead just pass in the exact server object?

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
spolti: moved to this function.

pkg/controller/kieapp/defaults/defaults.go
spolti: yeah, remove the extra func, it is cleaner now.

deploy/crs/v2/kieapp_rhdm_production-immutable.yaml
```diff
@@ -16,4 +16,9 @@ spec:
             uri: https://github.com/jboss-container-images/rhdm-7-openshift-image.git
             reference: master
             contextDir: quickstarts/hello-rules/hellorules
+          # uncomment the block below to set custom environment variable for BuildConfig, useful to configure JAVA_OPTS_APPEND or other
+          # properties such as MAVEN_OPTS_APPEND
+          # env:
+          #  - name: "JAVA_OPTS_APPEND"
+          #    value: "-Dmy.custom.property=prop-test"
```
tchughesiv, line 23: this seems like a good candidate for a snippet instead

tchughesiv (changes requested):

deploy/crs/v2/snippets/prior_version.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   annotations:
     consoleDesc: Use this snippet to deploy a prior product version
     consoleName: snippet-prior-version
-    consoleSnippet: "true"
+    consoleSnippet: true
```
tchughesiv, line 7: this change is unecessary

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -625,6 +627,12 @@ func ConsolidateObjects(env api.Environment, cr *api.KieApp) api.Environment {
 	for index := range env.Servers {
 		serverSet, _ := GetServerSet(cr, index)
 		env.Servers[index] = ConstructObject(env.Servers[index], serverSet.KieAppObject)
+		// apply the build config envs provided through cr to the given kieSerever BuildConfig Spec envs
+		for bcindex := range env.Servers[index].BuildConfigs {
+			env.Servers[index].BuildConfigs[bcindex].Spec.Strategy.SourceStrategy.Env = shared.EnvOverride(
+				env.Servers[index].BuildConfigs[bcindex].Spec.Strategy.SourceStrategy.Env,
+				cr.Spec.Objects.Servers[index].Build.Env)
```
tchughesiv, line 634: i believe this can be `serverSet.Build.Env`

tchughesiv (changes requested):

deploy/crs/v2/snippets/prior_version.yaml
spolti: oops, replacing back changed it by accident.

pkg/controller/kieapp/defaults/defaults.go
spolti: let me try.

deploy/crs/v2/kieapp_rhdm_production-immutable.yaml
spolti: alright, added to snippet.

tchughesiv (approved):

PR #533, opened 2020-10-27
[RHPAM-3245] - Change resource requirements for BC and KIE Server

Kaitou786:
Alongside the title JIRA this PR also solves the [BAPL-1766](https://issues.redhat.com/browse/BAPL-1766) as almost both were identical except for setting default value of `MaxMetaspaceSize`. Since that was a small thing I incorporated in this PR only
See: https://issues.redhat.com/browse/RHPAM-3245

Signed-off-by: Tarun Khandelwal <tarkhand@redhat.com>

PR #542, opened 2020-11-09
Remove default resource limits from template except for trial envs

Kaitou786:
Following the #533 the default resource limits set in the templates should be removed except  for the trial envs. In this PR the resource limits are moved from the common config to only for the trial envs

### DIFF
--- a/config/7.9.1/common.yaml
+++ b/config/7.9.1/common.yaml
@@ -40,9 +40,6 @@ console:
               - name: "[[.ApplicationName]]-[[.Console.Name]]"
                 image: "[[.Console.ImageURL]]"
                 imagePullPolicy: Always
-                resources:
-                  limits:
-                    memory: 2Gi
                 livenessProbe:
                   httpGet:
                     path: /rest/healthy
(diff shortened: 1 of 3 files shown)


config/7.9.1/common.yaml
```diff
@@ -40,9 +40,6 @@ console:
               - name: "[[.ApplicationName]]-[[.Console.Name]]"
                 image: "[[.Console.ImageURL]]"
                 imagePullPolicy: Always
-                resources:
-                  limits:
-                    memory: 2Gi
```
tchughesiv, line 45: is console the only component this need to be removed from? What about kieservers, smart router, & pim?

spolti: it should be remove on all places, except for the trial environment.

Kaitou786: Removed

spolti:
/lgtm

@tchughesiv will this changes for this be picked for 7.10 as well?

tchughesiv:
@spolti yes, these changes will be in 7.10 as well

PR #546, opened 2020-11-10
Revert GcMaxMetaspaceSize CRD to accept the int

Kaitou786:
Reverting as the cct_modules append a `m` in the env
For more info See: https://github.com/jboss-openshift/cct_module/blob/0.39.0/jboss/container/java/jvm/bash/artifacts/opt/jboss/container/java/jvm/java-default-options#L157

### DIFF
--- a/deploy/crds/kieapp.crd.yaml
+++ b/deploy/crds/kieapp.crd.yaml
@@ -535,9 +535,10 @@ spec:
                             format: int32
                             type: integer
                           gcMaxMetaspaceSize:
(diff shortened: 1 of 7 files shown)


spolti:
@tchughesiv sry for this PR.
we forgot that a external module that our image use appends the "m" at the end of the gcMaxMetaspaceSize, we would need to revert it back.

tchughesiv:
@spolti is this for 7.9.1?

spolti:
@tchughesiv yes.

PR #557, opened 2020-11-18
[KIECLOUD-438] - DeploymentConfig probes for BusinessCentral too shor…

Kaitou786:
…t causing POD Termination

See: https://issues.redhat.com/browse/KIECLOUD-438

### DIFF
--- a/config/7.9.1/common.yaml
+++ b/config/7.9.1/common.yaml
@@ -45,15 +45,25 @@ console:
                     path: /rest/healthy
                     port: 8080
                     scheme: HTTP
+                  initialDelaySeconds: 90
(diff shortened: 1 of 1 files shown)


spolti:
@Kaitou786 I think it needs to be cherry picked to master, right @tchughesiv ?

sutaakar:
This change causes **minimal** Business central startup time to be 3 minutes. That is unnecessary for fast environments with trial environment.
IMHO the probes deserve complex adjustments as IMHO with the startup probe the liveness probe doesn't need `initialDelaySeconds`, also thresholds could be adjusted.

PR #561, opened 2020-12-02
[RHPAM-3325] reconcile error due to mem conversion and startupProbe

tchughesiv:

### DIFF
--- a/pkg/controller/kieapp/constants/constants.go
+++ b/pkg/controller/kieapp/constants/constants.go
@@ -173,7 +173,7 @@ const (
 	ConsoleAuthoringCPURequests = "1500m"
 	ConsoleAuthoringMemRequests = "3Gi"
 	ConsoleProdCPURequests      = "500m"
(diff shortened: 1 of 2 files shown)


pkg/controller/kieapp/constants/constants.go
```diff
@@ -173,7 +173,7 @@ const (
 	ConsoleAuthoringCPURequests = "1500m"
 	ConsoleAuthoringMemRequests = "3Gi"
 	ConsoleProdCPURequests      = "500m"
-	ConsoleProdMemRequests      = "1.5Gi"
+	ConsoleProdMemRequests      = "1500m"
```
spolti, line 176: shouldn't it be 1500Mi or keep as 1.5Gi?

tchughesiv: We can be, I was just matching the other memory settings. Your call though

tchughesiv: done

PR #568, opened 2020-12-09
KIECLOUD-480 - Allowing value flag as "optional" for LdapExtended log…

msmagnanijr:
… in module

See: https://issues.redhat.com/browse/KIECLOUD-480

Signed-off-by: Mauricio Magnani <msmagnanijr@gmail.com>

### DIFF
--- a/config/7.10.0/common.yaml
+++ b/config/7.10.0/common.yaml
@@ -159,6 +159,8 @@ console:
                     value: "[[.Auth.LDAP.BindCredential]]"
                   - name: AUTH_LDAP_JAAS_SECURITY_DOMAIN
                     value: "[[.Auth.LDAP.JAASSecurityDomain]]"
+                  - name: AUTH_LDAP_LOGIN_MODULE
+                    value: "[[.Auth.LDAP.LoginModule]]"
                   - name: AUTH_LDAP_BASE_CTX_DN
                     value: "[[.Auth.LDAP.BaseCtxDN]]"
                   - name: AUTH_LDAP_BASE_FILTER
--- a/deploy/crds/kieapp.crd.yaml
+++ b/deploy/crds/kieapp.crd.yaml
@@ -86,6 +86,13 @@ spec:
                         description: The JMX ObjectName of the JaasSecurityDomain
                           used to decrypt the password.
                         type: string
+                      loginModule:
+                        description: A flag to set login module to optional. The default
+                          value is required
+                        enum:
+                        - optional
+                        - required
+                        type: string
                       parseRoleNameFromDN:
                         description: A flag indicating if the DN returned by a query
                           contains the roleNameAttributeID. If set to true, the DN
@@ -1772,6 +1779,13 @@ spec:
                           jaasSecurityDomain:
                             description: The JMX ObjectName of the JaasSecurityDomain
                               used to decrypt the password.
+                            type: string
+                          loginModule:
+                            description: A flag to set login module to optional. The
+                              default value is required
+                            enum:
+                            - optional
+                            - required
                             type: string
                           parseRoleNameFromDN:
                             description: A flag indicating if the DN returned by a
--- a/deploy/ui/form.json
+++ b/deploy/ui/form.json
@@ -247,6 +247,13 @@
                   "description": "LDAP Credentials used for authentication of bindDN"
                 },
                 {
+                  "label": "LoginModule Flag",
+                  "type": "dropDown",
+                  "jsonPath": "$.spec.auth.ldap.loginModule",
+                  "originalJsonPath": "$.spec.auth.ldap.loginModule",
+                  "description": "A flag to set login module to optional. The default value is required"
+                },
+                {
                   "label": "JAAS security domain",
                   "type": "password",
                   "jsonPath": "$.spec.auth.ldap.jaasSecurityDomain",
--- a/pkg/apis/app/v2/kieapp_types.go
+++ b/pkg/apis/app/v2/kieapp_types.go
@@ -356,6 +356,8 @@ type LDAPAuthConfig struct {
 	BindDN string `json:"bindDN,omitempty"`
 	// The JMX ObjectName of the JaasSecurityDomain used to decrypt the password.
 	JAASSecurityDomain string `json:"jaasSecurityDomain,omitempty"`
+	// +kubebuilder:validation:Enum:=optional;required
+	LoginModule LoginModuleType `json:"loginModule,omitempty"`
 	// LDAP Base DN of the top-level context to begin the user search.
 	BaseCtxDN string `json:"baseCtxDN,omitempty"`
 	// DAP search filter used to locate the context of the user to authenticate. The input username or userDN obtained from the login module callback is substituted into the filter anywhere a {0} expression is used. A common example for the search filter is (uid={0}).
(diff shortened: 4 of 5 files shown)


msmagnanijr:
@spolti could you please take a look? thanks!

tchughesiv (changes requested):
Please run a `make test` and commit the changes. Also, can you add a unit test for this change?

spolti:
@msmagnanijr what is the status of this issue?

msmagnanijr:
> @msmagnanijr what is the status of this issue?

I created the test but I'm not sure if it went as it should.

pkg/controller/kieapp/defaults/auth_test.go
```diff
@@ -585,3 +585,36 @@ func getExpectedSSOEnvs() []corev1.EnvVar {
...
+				LDAP: &api.LDAPAuthConfig{
+					LoginModule: "required",
+				},
+			},
+		},
+	}
+	env, err := GetEnvironment(cr, test.MockService())
+	assert.Nil(t, err, "Error getting trial environment")
+
+	expectedEnvs := []corev1.EnvVar{
```
spolti, line 611: there is a easier way to check envs:

eg:

```
assert.Equal(t, "dataset_1,dataset_2", getEnvVariable(env.Dashbuilder.DeploymentConfigs[0].Spec.Template.Spec.Containers[0], "KIESERVER_DATASETS"))
```

spolti (changes requested):

pkg/controller/kieapp/defaults/auth_test.go
```diff
@@ -585,3 +585,36 @@ func getExpectedSSOEnvs() []corev1.EnvVar {
 		{Name: "SSO_PASSWORD"},
 	}
 }
+
+func testLDAPFlag(t *testing.T) {
```
spolti, line 589:
```suggestion
func testLDAPLoginModuleFlag(t *testing.T) {
```

spolti (changes requested):

pkg/controller/kieapp/defaults/auth_test.go
```diff
@@ -585,3 +585,29 @@ func getExpectedSSOEnvs() []corev1.EnvVar {
...
+				LDAP: &api.LDAPAuthConfig{
+					LoginModule: "required",
+				},
+			},
+		},
+	}
+	env, err := GetEnvironment(cr, test.MockService())
+	assert.Nil(t, err, "Error getting trial environment")
+
+	assert.Equal(t, "required", getEnvVariable(env.Console.DeploymentConfigs[0].Spec.Template.Spec.Containers[0], "AUTH_LDAP_LOGIN_MODULE"))
```
tchughesiv, line 611: can you add another test that checks when set to optional?

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/auth_test.go
msmagnanijr: done!

spolti: seems you forgot to test this new property on the the Servers' object as well

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -356,6 +356,8 @@ type LDAPAuthConfig struct {
 	BindDN string `json:"bindDN,omitempty"`
 	// The JMX ObjectName of the JaasSecurityDomain used to decrypt the password.
 	JAASSecurityDomain string `json:"jaasSecurityDomain,omitempty"`
+	// A flag to set login module to optional. The default value is required
+	LoginModule string `json:"loginModule,omitempty"`
```
tchughesiv, line 360: are `required` and `optional` the only 2 possibilities here? if so, maybe we add them as enum options instead?
https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/apis/app/v2/kieapp_types.go#L19

config/7.10.0/common.yaml
```diff
@@ -159,6 +159,8 @@ console:
                     value: "[[.Auth.LDAP.BindCredential]]"
                   - name: AUTH_LDAP_JAAS_SECURITY_DOMAIN
                     value: "[[.Auth.LDAP.JAASSecurityDomain]]"
+                  - name: AUTH_LDAP_LOGIN_MODULE
+                    value: "[[.Auth.LDAP.LoginModule]]"
```
tchughesiv, line 163: what if nothing is set and the resulting `AUTH_LDAP_LOGIN_MODULE` var is blank? is this ok or should we be assigning a default value via code?

spolti: the images set a default value if it is empty.

pkg/apis/app/v2/kieapp_types.go
spolti: yes, only two options, this is actually a good idea. :+1:

deploy/crds/kieapp.crd.yaml
```diff
@@ -86,6 +86,10 @@ spec:
                         description: The JMX ObjectName of the JaasSecurityDomain
                           used to decrypt the password.
                         type: string
+                      loginModule:
+                        description: A flag to set login module to optional. The default
+                          value is required
+                        type: string
```
spolti, line 92: to make the field as enum you need to do as it is done here:
https://github.com/kiegroup/kie-cloud-operator/blob/master/pkg/apis/app/v2/kieapp_types.go#L21

set the var type and on its comments, add the enum valid values.

@tchughesiv  was it that you have suggested?

spolti (changes requested):

deploy/crds/kieapp.crd.yaml
msmagnanijr: @spolti I added the enum here but I really forgot to update the type in crd.  https://github.com/kiegroup/kie-cloud-operator/commit/448a87c671a4c54170ed54142d028342bee8a118#diff-4d98202026f495310fe4aa8f64f71057eb9fa83104da7480469153b7d6fe810bR360

msmagnanijr: I guess it could be something like this:

```
                      searchScope:
                        description: SearchScopeType Type used to define how the LDAP
                          searches are performed
                        enum:
                        - SUBTREE_SCOPE
                        - OBJECT_SCOPE
                        - ONELEVEL_SCOPE
                        type: string
```

spolti: the crd is automatically updated, no manual changes are required there.

tchughesiv (approved):

spolti (approved):

deploy/ui/form.json
```diff
@@ -246,6 +246,12 @@
                   "jsonPath": "$.spec.auth.ldap.bindCredential",
                   "description": "LDAP Credentials used for authentication of bindDN"
                 },
+                {
+                  "label": "LoginModule Flag",
+                  "type": "text",
+                  "jsonPath": "$.spec.auth.ldap.loginModule",
+                  "description": "A flag to set login module to optional. The default value is required"
```
tchughesiv, line 253: should the type be set to `dropDown` as seen here?
https://github.com/kiegroup/kie-cloud-operator/blob/8d35c107d0d110be689de7978f8af237a9b7cfeb/deploy/ui/form.json#-L18-L20

tchughesiv (changes requested):

tchughesiv (approved):

PR #572, opened 2020-12-17
[RHPAM-3320] - Business Central PV size is not configurable in Operator

spolti:

### DIFF
--- a/pkg/apis/app/v2/kieapp_types.go
+++ b/pkg/apis/app/v2/kieapp_types.go
@@ -134,6 +134,7 @@ type ConsoleObject struct {
 	SSOClient    *SSOAuthClient  `json:"ssoClient,omitempty"`
 	GitHooks     *GitHooksVolume `json:"gitHooks,omitempty"`
 	Jvm          *JvmObject      `json:"jvm,omitempty"`
+	PvSize       string          `json:"pvSize,omitempty"`
 }
 
 // SmartRouterObject configuration of the RHPAM smart router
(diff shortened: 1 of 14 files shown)


spolti:
Hey @tchughesiv , do you think we can review it and try to get it merged this week?

bmozaffa:
@spolti Almost everyone on our team, including Tommy, are off this week. If there is urgency around this, I can review and catch up, and get it merged. If there isn't, best to wait for Tommy on Jan 4. Let me know!

Kaitou786:
Code, looks ok to me, just one thing I don't think there should be any change in CSV file, can you revert them back @spolti ?

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -240,6 +240,8 @@ type KieAppObject struct {
 	ImageTag string `json:"imageTag,omitempty"`
 	// The storageClassName to use
 	StorageClassName string `json:"storageClassName,omitempty"`
+	// The Size of the PV
+	PvSize string `json:"pvSize,omitempty"`
```
tchughesiv, line 244: If this is something that should only be configurable for the Console deployment, we may want to move it to the `ConsoleObject` struct instead. This `KieAppObject` is for common fields used by all object types... servers, smartrouters, etc. as well.

If instead you believe this should stay where it is, we should add support for the other object types as well in our templates. Otherwise, it will confuse the user to see it configurable in the server / router schema if it doesn't do anything.

tchughesiv (changes requested):

pkg/apis/app/v2/kieapp_types.go
spolti: updated  it.

tchughesiv (approved):

PR #573, opened 2020-12-21
[RHPAM-3319] - hard to deploy immutable kie-server with pre-built kja…

Kaitou786:
…r by KieApp installer

See: https://issues.redhat.com/browse/RHPAM-3319

### DIFF
--- a/config/7.10.1/envs/rhdm-production-immutable.yaml
+++ b/config/7.10.1/envs/rhdm-production-immutable.yaml
@@ -32,7 +32,7 @@ servers:
 
       ## KIE server deployment config END
     ## KIE server build config BEGIN
-    #[[if .Build.KieServerContainerDeployment]]
+    #[[if .Build.GitSource.URI]]
     imageStreams:
       - metadata:
           name: "[[.KieName]]"
--- a/config/7.10.1/envs/rhpam-production-immutable.yaml
+++ b/config/7.10.1/envs/rhpam-production-immutable.yaml
@@ -33,7 +33,7 @@ servers:
                       value: "http"
       ## KIE server deployment config END
     ## KIE server build config BEGIN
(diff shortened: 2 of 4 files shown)


deploy/crds/kieapp.crd.yaml
```diff
@@ -1471,6 +1468,9 @@ spec:
                         keystoreSecret:
                           description: Keystore secret name
                           type: string
+                        kieServerContainerDeployment:
+                          description: The Maven GAV to deploy, e.g., rhpam-kieserver-library=org.openshift.quickstarts:rhpam-kieserver-library:1.5.0-SNAPSHOT
+                          type: string
```
jakubschwan, line 1473: @tchughesiv is this api change safe from point of auto updates? `kieServerContainerDeployment` is moved from server.Build object to server. I'm little worried that this change could broke down existing KieApps during updates.

@Kaitou786 As the issue is that s2i build is execute when not needed (and git properties are required), maybe it could be enough to not require git properties (make them optional, remove 'require' flag from api) and do not trigger S2I build when there is set only container GAV.

tchughesiv: correct @jakubschwan ... this change would break things and cannot be done without an api version increase. this would complicate the change significantly as upgrade logic would have to be added in the form of a conversion webhook.

only api additions are allowed without versioning the api.

Kaitou786:
@jakubschwan can you look again, Now the api and UI form is same as previously but the BC will be created only when GIT properties are set

spolti (approved):

spolti:
/lgtm

nice work @Kaitou786

tchughesiv (changes requested):
please move these config changes to the 7.10.1 configs instead

tchughesiv (approved):

PR #574, opened 2020-12-21
[KIECLOUD-480] - Execute and Fix Unit Tests

Kaitou786:
After merging the KIECLOUD-480 [PR](https://github.com/kiegroup/kie-cloud-operator/pull/568) the unit test added were not getting executed. This PR fixes them and executes them from `go test`

### DIFF
--- a/pkg/controller/kieapp/defaults/auth_test.go
+++ b/pkg/controller/kieapp/defaults/auth_test.go
@@ -586,13 +586,13 @@ func getExpectedSSOEnvs() []corev1.EnvVar {
 	}
 }
 
-func testLDAPLoginModuleRequiredFlag(t *testing.T) {
+func TestLDAPLoginModuleRequiredFlag(t *testing.T) {
 	cr := &api.KieApp{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "test",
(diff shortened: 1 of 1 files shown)


pkg/controller/kieapp/defaults/auth_test.go
```diff
@@ -595,37 +596,42 @@ func testLDAPLoginModuleRequiredFlag(t *testing.T) {
...
 				LDAP: &api.LDAPAuthConfig{
+					URL:         "ldaps://ldap.example.com",
+					BindDN:      "cn=admin,dc=example,dc=com",
 					LoginModule: "required",
 				},
 			},
 		},
 	}
 	env, err := GetEnvironment(cr, test.MockService())
+	fmt.Println(err)
```
spolti, line 612: it will thrown a null pointer if err is nil, can be removed.

spolti (changes requested):

pkg/controller/kieapp/defaults/auth_test.go
Kaitou786: Ah yes, I forgot to remove this, added for debug purposes :p

PR #588, opened 2021-02-05
[RHPAM-3420] - Wrong image pulled for rhdm-production-immutable envir…

Kaitou786:
…onment

See: https://issues.redhat.com/browse/RHPAM-3420

### DIFF
--- a/pkg/controller/kieapp/defaults/merge_test.go
+++ b/pkg/controller/kieapp/defaults/merge_test.go
@@ -289,7 +289,6 @@ func TestMergeBuildConfigandIStreams(t *testing.T) {
 	assert.Equal(t, "test-kieserver", server.BuildConfigs[0].ObjectMeta.Name)
 	assert.Empty(t, server.DeploymentConfigs[0].Spec.Triggers[0].ImageChangeParams.From.Namespace)
 	assert.Equal(t, "test-kieserver:latest", server.DeploymentConfigs[0].Spec.Triggers[0].ImageChangeParams.From.Name)
-	assert.Equal(t, "test-kieserver", server.DeploymentConfigs[0].Spec.Template.Spec.Containers[0].Image)
 }
 
 func TestMergeDeploymentconfigs(t *testing.T) {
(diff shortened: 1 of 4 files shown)


spolti (approved):
Looks ok to me, spoke with @Kaitou786, just need to run some tests.
Approving, but let's wait his response to merge.

tchughesiv:
/hold

@Kaitou786 has this been thoroughly tested manually in a cluster?

Kaitou786:
> /hold
> 
> @Kaitou786 has this been thoroughly tested manually in a cluster?

I ran the tests @spolti asked me to run on the cluster, they seem to run fine

pkg/controller/kieapp/defaults/merge_test.go
```diff
@@ -289,7 +289,6 @@ func TestMergeBuildConfigandIStreams(t *testing.T) {
 	assert.Equal(t, "test-kieserver", server.BuildConfigs[0].ObjectMeta.Name)
 	assert.Empty(t, server.DeploymentConfigs[0].Spec.Triggers[0].ImageChangeParams.From.Namespace)
 	assert.Equal(t, "test-kieserver:latest", server.DeploymentConfigs[0].Spec.Triggers[0].ImageChangeParams.From.Name)
-	assert.Equal(t, "test-kieserver", server.DeploymentConfigs[0].Spec.Template.Spec.Containers[0].Image)
```
tchughesiv, line 292: instead of removing this, is there a different check we should be running instead? also, do we have coverage for these immutable envs when a build is NOT configured?

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/merge_test.go
Kaitou786: Since without build we won't have bc or ImageStream, so I don't think there is any alternative for this particular test, but we add some unit test for the production immutable env, I added the image verification as part of that change.

tchughesiv: the kie server has to use an image, otherwise we have a failed deployment.. or a DC just sitting there. does an immutable env not have a default image when BC not used?

Kaitou786: That's what I defined in the test. The default image for kieserver when BC is not defined
It should be registry.redhat.io/rhpam-7/rhpam-kieserver. Does it need to be something else?

PR #596, opened 2021-02-18
Merge branch 'main' into release-v7.10.x

tchughesiv:
version 7.10.1


spolti:
seems that the error is related with some infra problem?
/approve

tchughesiv:
the one error is related to configs ... this error is only allowed when doing this initial mass merge like this. as we decrease how many product versions each operator support, this check may go away.

PR #603, opened 2021-02-19
[RHPAM-3471] 7.10.x Label for subcomp is wrongly set in Operator for images with digest

desmax74:
See https://issues.redhat.com/browse/RHPAM-3471

@tchughesiv

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -5,6 +5,7 @@ package defaults
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -565,7 +589,15 @@ func GetImage(imageURL string) (image, imageTag, imageContext string) {
 	if len(imageParts) > 1 {
 		imageTag = imageParts[len(imageParts)-1]
 	}
-	return image, imageTag, imageContext
+	return image, imageTag, imageContext, isImageValid(imageAndTag)
+}
+
+func isImageValid(imageAndTag string) error {
```
spolti, line 595: what is the purpose of this function if all places the GetImage func is called the error is ignored?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -139,13 +140,13 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 }
 
 func setProductLabels(cr *api.KieApp, env *api.Environment) {
-	setObjectLabels(cr, &env.Console)
-	setObjectLabels(cr, &env.Dashbuilder)
+	setObjectLabels(cr, &env.Console, getFormattedComponentName(cr, "businesscentral"))
```
spolti, line 143: does not hardcode values here, businesscentral is different than decision central, this value should come from env.Console.

same for the hardcoded values below.

spolti (changes requested):

PR #604, opened 2021-02-19
[RHPAM-3481] Label for subcomp is wrongly set in Operator for images with digest

desmax74:
See https://issues.redhat.com/browse/RHPAM-3481

@tchughesiv

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -139,35 +139,71 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 }
 
 func setProductLabels(cr *api.KieApp, env *api.Environment) {
-	setObjectLabels(cr, &env.Console)
-	setObjectLabels(cr, &env.Dashbuilder)
+	setObjectLabels(cr, &env.Console, getFormattedComponentName(cr, getConsoleName(cr)))
+	setObjectLabels(cr, &env.Dashbuilder, getFormattedComponentName(cr, constants.DashBuilder))
 	for index := range env.Servers {
-		setObjectLabels(cr, &env.Servers[index])
+		setObjectLabelsForServer(cr, &env.Servers[index])
 	}
-	setObjectLabels(cr, &env.SmartRouter)
-	setObjectLabels(cr, &env.ProcessMigration)
+	setObjectLabels(cr, &env.SmartRouter, getFormattedComponentName(cr, constants.Smartrouter))
+	setObjectLabels(cr, &env.ProcessMigration, getFormattedComponentName(cr, constants.ProcessMigration))
+}
+
+func getConsoleName(cr *api.KieApp) string {
+	env := string(cr.Spec.Environment)
+	consoleName := ""
+	if strings.Contains(env, constants.RhdmPrefix) {
+		consoleName = constants.RhdmDecisionCentral
+	} else if strings.Contains(env, constants.RhpamPrefix) {
+
+		if strings.Contains(env, constants.Production) {
+			consoleName = constants.RhpamBusinessCentralMon
+		} else {
+			consoleName = constants.RhpamBusinessCentral
+		}
+	}
+	return consoleName
 }
 
 func getSubComponentTypeByImageName(imageName string) string {
-	retValue := "application"
+	retValue := constants.SUBCOMPONENT_TYPE_APP
 	if imageName == constants.RhpamSmartRouterImageName || imageName == constants.RhpamControllerImageName ||
 		imageName == constants.RhdmSmartRouterImageName || imageName == constants.RhdmControllerImageName {
 
-		retValue = "infrastructure"
+		retValue = constants.SUBCOMPONENT_TYPE_INFRA
 	}
 	return retValue
 }
 
-func setObjectLabels(cr *api.KieApp, object *api.CustomObject) {
+func setObjectLabels(cr *api.KieApp, object *api.CustomObject, subcomponent string) {
 	for index, obj := range object.DeploymentConfigs {
-		subcomponent, _, _ := GetImage(object.DeploymentConfigs[index].Spec.Template.Spec.Containers[0].Image)
 		object.DeploymentConfigs[index].Spec.Template.Labels = setLabels(cr, obj.Spec.Template.Labels, subcomponent, getSubComponentTypeByImageName(subcomponent))
 	}
 	for index, obj := range object.StatefulSets {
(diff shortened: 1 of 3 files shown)


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -158,16 +159,39 @@ func getSubComponentTypeByImageName(imageName string) string {
...
 	}
 	for index, obj := range object.StatefulSets {
-		subcomponent, _, _ := GetImage(object.StatefulSets[index].Spec.Template.Spec.Containers[0].Image)
 		object.StatefulSets[index].Spec.Template.Labels = setLabels(cr, obj.Spec.Template.Labels, subcomponent, getSubComponentTypeByImageName(subcomponent))
 	}
+}
+
+func setObjectLabelsForServer(cr *api.KieApp, object *api.CustomObject) {
+	for index, obj := range object.DeploymentConfigs {
+		subcomponent := getFormattedComponentName(cr, object.DeploymentConfigs[index].Name)
```
tchughesiv, line 173: this won't work, as a user can name their kieservers however they like. a user could name their kieserver `foo` if they like. instead, something like this might work?
```golang
		subcomponent := getFormattedComponentName(cr, "kieserver")
```

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -554,7 +578,7 @@ func getSmartRouterTemplate(cr *api.KieApp) api.SmartRouterTemplate {
 }
 
 // GetImage ...
-func GetImage(imageURL string) (image, imageTag, imageContext string) {
+func GetImage(imageURL string) (image, imageTag, imageContext string, error error) {
```
tchughesiv, line 581: instead of error, should we use a `bool` here? another option i've considered is to modify the function to properly return image/imagetag even if the image is using a digest. so the tag returned would be the proper `sha256:xxxxxxxxcb244ea5f127f28d36fa88069f864f54xxxxx` if an `@` is detected

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -139,13 +140,13 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 }
 
 func setProductLabels(cr *api.KieApp, env *api.Environment) {
-	setObjectLabels(cr, &env.Console)
-	setObjectLabels(cr, &env.Dashbuilder)
+	setObjectLabels(cr, &env.Console, getFormattedComponentName(cr, "businesscentral"))
```
spolti, line 143: does not hardcode values here, businesscentral is different than decision central, this value should come from env.Console.

same for the hardcoded values below.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -565,7 +589,15 @@ func GetImage(imageURL string) (image, imageTag, imageContext string) {
 	if len(imageParts) > 1 {
 		imageTag = imageParts[len(imageParts)-1]
 	}
-	return image, imageTag, imageContext
+	return image, imageTag, imageContext, isImageValid(imageAndTag)
+}
+
+func isImageValid(imageAndTag string) error {
```
spolti, line 595: what is the purpose of this function if all places the GetImage func is called the error is ignored?

spolti (changes requested):

pkg/controller/kieapp/defaults/defaults.go
desmax74: Ok, which field we want to use from env.Console ?

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: also, given that labels will no longer be leveraging this `GetImage` function... any changes to this function should probably be done separately. please move changes specific to this function to a new PR.

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
desmax74: I'll change in a different PR

pkg/controller/kieapp/defaults/defaults.go
desmax74: Removed for another PR

pkg/controller/kieapp/defaults/defaults.go
desmax74: Changed

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -139,35 +139,74 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+
+func getFormattedComponentName(cr *api.KieApp, name string) string {
+	return getPrefixEnv(cr) + "-" + name + constants.RhelVersion
+}
+
+func getPrefixEnv(cr *api.KieApp) string {
+	prefix := ""
+	if strings.HasPrefix(string(cr.Status.Applied.Environment), constants.RhpamPrefix) {
+		prefix = constants.RhpamPrefix
+	}
```
tchughesiv, line 205: would it make more sense to set prefix to `constants.RhpamPrefix` by default, and only do the strings check for the rhdm prefix?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -139,35 +139,71 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+		consoleName = constants.RhdmDecisionCentral
+	} else if strings.Contains(env, constants.RhpamPrefix) {
+
+		if strings.Contains(env, constants.Production) {
+			consoleName = constants.RhpamBusinessCentralMon
+		} else {
+			consoleName = constants.RhpamBusinessCentral
+		}
+	}
+	return consoleName
```
tchughesiv, line 164: do we have unit test coverage for each of these possibilities?

tchughesiv:
@desmax74 `main` branch is not set to 7.11.0 code base. i've switched this PR to target `release-7.10.x`. once merged, we'll need to cherry-pick to `main` as well.

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: seems `constants.RhpamBusinessCentral` is being tested but not the other console possibilities?

tchughesiv (approved):

PR #613, opened 2021-03-15
 [RHPAM-3330-str] Operator/Image support for Kie-Server Cluster support 

desmax74:
PR to test console-cr-form with jbpm as a string in a dropdown list 
https://issues.redhat.com/browse/KIECLOUD-507

### DIFF
--- a/config/7.11.0/common.yaml
+++ b/config/7.11.0/common.yaml
@@ -680,6 +680,8 @@ servers:
                       value: "[[.Build.DisablePullDeps]]"
                     - name: KIE_SERVER_DISABLE_KC_VERIFICATION
                       value: "[[.Build.DisableKCVerification]]"
+                    - name: JBPM_CLUSTER
+                      value: "true"
                     # Auth config BEGIN
                     ## SSO config BEGIN
                     #[[if $.Auth.SSO.URL]]
--- a/pkg/apis/app/v2/kieapp_types.go
+++ b/pkg/apis/app/v2/kieapp_types.go
@@ -252,6 +252,8 @@ type KieAppObject struct {
 	ImageTag string `json:"imageTag,omitempty"`
 	// The storageClassName to use
 	StorageClassName string `json:"storageClassName,omitempty"`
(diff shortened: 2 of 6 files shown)


pkg/apis/app/v2/kieapp_types.go
```diff
@@ -252,6 +252,8 @@ type KieAppObject struct {
 	ImageTag string `json:"imageTag,omitempty"`
 	// The storageClassName to use
 	StorageClassName string `json:"storageClassName,omitempty"`
+	// Enable the clustering between Kieservers and add one replicas if only one is configured
+	JbpmCluster string `json:"jbpmCluster,omitempty"`
```
tchughesiv, line 256: shouldn't this be a bool instead?

config/7.11.0/common.yaml
```diff
@@ -680,6 +680,8 @@ servers:
                       value: "[[.Build.DisablePullDeps]]"
                     - name: KIE_SERVER_DISABLE_KC_VERIFICATION
                       value: "[[.Build.DisableKCVerification]]"
+                    - name: JBPM_CLUSTER
+                      value: "true"
```
tchughesiv, line 684: if this is set statically to true, how would the user change to false? maybe this needs to be a template var instead?
`[[.JbpmCluster]]` ?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -695,6 +696,19 @@ func getServersConfig(cr *api.KieApp) ([]api.ServerTemplate, error) {
 			}
 
 			// Set replicas
+			// if the configuration use only one replica we increment
+			if len(cr.Spec.Objects.Servers) != 0 {
```
tchughesiv, line 700: at this point in the code, in this servers loop, we already know len is not 0. are we sure this check is necessary?

tchughesiv (changes requested):

desmax74:
@tchughesiv thanks for the review but this was just for a test with swati

PR #614, opened 2021-03-15
[RHPAM-3331] Operator for AMQ Streams (Kafka) integration

desmax74:
See https://issues.redhat.com/browse/RHPAM-3331


config/7.11.0/common.yaml
```diff
@@ -829,6 +829,40 @@ servers:
...
+                    #[[end]]
+                    #[[if .KafkaExt.Acks]]
+                    - name: org.kie.server.jbpm-kafka.ext.acks
+                      value: "[[.KafkaExt.Acks]]"
+                    #[[end]]
+                    #[[if .KafkaExt.MaxBlockMs]]
+                    - name: org.kie.server.jbpm-kafka.ext.max.block.ms
+                      value: "[[.KafkaExt.MaxBlockMs]]"
+                    #[[end]]
+                    ## AMQ Streams END
```
jakubschwan, line 865: Env vars like these are not exists on Kie Server image, this configuration will not be passed correctly to the Kie Server image. 
On Kie Server image must be created env vars, witch can configured needed system properties. Then in Kie Operator we can pass needed values to the correct Kie Server's env vars.

deploy/crds/kieapp.crd.yaml
```diff
@@ -1871,6 +1871,41 @@ spec:
                                 to generated options in JAVA_OPTS. e.g. '-Dsome.property=foo'
                               type: string
                           type: object
+                        kafkaExt:
+                          description: KafkaExt kafka configuration to be used by
```
jakubschwan, line 1875: IMO It would be better to use there Kafka integration naming then extension, as this configuration should server for Kie Server and Kafka integration.

deploy/crds/kieapp.crd.yaml
```diff
@@ -1871,6 +1871,41 @@ spec:
...
+                            bootstrapServers:
+                              description: A comma separated list of bootstrapServers
+                              type: string
+                            clientID:
+                              description: Identifier to pass to the server when making
+                                requests
+                              type: string
+                            extDisabled:
+                              description: Disable the Kafka Extension
+                              type: string
```
jakubschwan, line 1895: Is this needed? We users should disabled configured kafka? If Kafka integration is not configured, there is no need to disabled it.

jakubschwan, line 1895: if this is needed, wouldn't be better for this use type boolean

deploy/crds/kieapp.crd.yaml
```diff
@@ -1871,6 +1871,41 @@ spec:
...
+                            groupID:
+                              description: Group identifier the group this consumer
+                                belongs
+                              type: string
+                            maxBlockMs:
+                              description: Number of milliseconds that indicates how
+                                long publish method will block
+                              type: string
```
jakubschwan, line 1903: as this is number, an integer would be better type for this parameter

jakubschwan (changes requested):

config/7.11.0/common.yaml
spolti: I was about to ask for the kieserver image changes.

deploy/crds/kieapp.crd.yaml
spolti: I guess, it should be exposed only on the ui, if enabled expose the other kafka parameters just like we have for AMQ configuration.

deploy/crds/kieapp.crd.yaml
spolti: ms maybe make more sense to use int64, right?

config/7.11.0/common.yaml
```diff
@@ -829,6 +829,40 @@ servers:
                       value: "[[.Jvm.GcContainerOptions]]"
                     #[[end]]
                     ## Jvm config END
+                    ## AMQ Streams BEGIN
+                    #[[if .KafkaExt.ExtDisabled]]
```
spolti, line 833: is it really needed? or is it just a control flag?
Is it a conf that the kieserevr image will expect?

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -942,6 +944,26 @@ type ObjectReference struct {
 	FieldPath string `json:"fieldPath,omitempty" protobuf:"bytes,7,opt,name=fieldPath"`
 }
 
+// KafkaExt kafka configuration to be used by the KieApp
+type KafkaExt struct {
+	// Disable the Kafka Extension
+	ExtDisabled string `json:"extDisabled,omitempty"`
```
spolti, line 950: should be *bool

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -942,6 +944,26 @@ type ObjectReference struct {
...
+	// A comma separated list of pairs <message/signal name>=<topicName>
+	Topics string `json:"topics,omitempty"`
```
spolti, line 952: would be better have a list instead a string here?
then during env provisioning you read this env and properly configure it.

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -942,6 +944,26 @@ type ObjectReference struct {
...
+	// Identifier to pass to the server when making requests
+	ClientID string `json:"clientID,omitempty"`
+	// Allow automatic topic creation.
+	AutocreateTopics string `json:"autocreateTopics,omitempty"`
+	// Group identifier the group this consumer belongs
+	GroupID string `json:"groupID,omitempty"`
+	// The number of acknowledgments the producer requires the leader to have received before considering a request complete.
+	Acks string `json:"acks,omitempty"`
+	// Number of milliseconds that indicates how long publish method will block
+	MaxBlockMs string `json:"maxBlockMs,omitempty"`
```
spolti, line 964: int64

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -942,6 +944,26 @@ type ObjectReference struct {
...
+	// A comma separated list of bootstrapServers
+	BootstrapServers string `json:"bootstrapServers,omitempty"`
+	// Identifier to pass to the server when making requests
+	ClientID string `json:"clientID,omitempty"`
+	// Allow automatic topic creation.
+	AutocreateTopics string `json:"autocreateTopics,omitempty"`
+	// Group identifier the group this consumer belongs
+	GroupID string `json:"groupID,omitempty"`
+	// The number of acknowledgments the producer requires the leader to have received before considering a request complete.
+	Acks string `json:"acks,omitempty"`
```
spolti, line 962: int

pkg/controller/kieapp/constants/constants.go
```diff
@@ -260,6 +260,26 @@ const (
 	//Envs
 	KubeNS     = "KUBERNETES_NAMESPACE"
 	KubeLabels = "KUBERNETES_LABELS"
+
+	//Kafka extension
+	KIE_KAFKA_EXT_DISABLED         = "org.kie.kafka.server.ext.disabled"
```
spolti, line 265: These envs that sets the properties would need to go on the image, yes?
@jakubschwan @tchughesiv wdyt?

spolti (changes requested):

pkg/controller/kieapp/constants/constants.go
jakubschwan: yes, looks like these should go to the image. I would left there only constants for default values if needed

deploy/crds/kieapp.crd.yaml
desmax74: They used this naming

deploy/crds/kieapp.crd.yaml
desmax74: By default is disabled

desmax74: Yep, was string due to the problem with UI (checkbox)

deploy/crds/kieapp.crd.yaml
desmax74: converted to int, if they need huge amount of time to block we'll increase

config/7.11.0/common.yaml
desmax74: ok

pkg/apis/app/v2/kieapp_types.go
desmax74: yep

pkg/apis/app/v2/kieapp_types.go
desmax74: No, because they wait for a string with all topics configurations pairs, is up to the java code splitting with comma

pkg/apis/app/v2/kieapp_types.go
desmax74: int is better to start, are milliseconds to express seconds to block

pkg/apis/app/v2/kieapp_types.go
desmax74: yep

pkg/controller/kieapp/constants/constants.go
desmax74: jboss-container-images/rhdm-7-openshift-image#413
jboss-container-images/rhpam-7-openshift-image#524

pkg/controller/kieapp/constants/constants.go
```diff
@@ -260,6 +260,23 @@ const (
 	//Envs
 	KubeNS     = "KUBERNETES_NAMESPACE"
 	KubeLabels = "KUBERNETES_LABELS"
+
+	//Kafka extension
+	KIE_SERVER_KAFKA_EXT_DISABLED          = "KIE_SERVER_KAFKA_EXT_DISABLED"
```
spolti, line 265: Not sure if it is needed here, besides the tests, what are the other use case?

pkg/controller/kieapp/constants/constants.go
```diff
@@ -260,6 +260,23 @@ const (
...
+	KIE_SERVER_KAFKA_EXT_TOPICS            = "KIE_SERVER_KAFKA_EXT_TOPICS"
+	KIE_SERVER_KAFKA_EXT_SERVERS           = "KIE_SERVER_KAFKA_EXT_SERVERS"
+	KIE_SERVER_KAFKA_EXT_CLIENT_ID         = "KIE_SERVER_KAFKA_EXT_CLIENT_ID"
+	KIE_SERVER_KAFKA_EXT_AUTOCREATE_TOPICS = "KIE_SERVER_KAFKA_EXT_AUTOCREATE_TOPICS"
+	KIE_SERVER_KAFKA_EXT_GROUP_ID          = "KIE_SERVER_KAFKA_EXT_GROUP_ID"
+	KIE_SERVER_KAFKA_EXT_ACKS              = "KIE_SERVER_KAFKA_EXT_ACKS"
+	KIE_SERVER_KAFKA_EXT_MAX_BLOCK_MS      = "KIE_SERVER_KAFKA_EXT_MAX_BLOCK_MS"
+
+	//default values
+	KIE_KAFKA_EXT_MAX_BLOCK_MS_DEFAULT      = 2000
```
spolti, line 275: I think the pattern for consts in Golang is camelcase.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1293,6 +1299,28 @@ func setJvmDefault(jvm *api.JvmObject) {
...
+			//if aren't present bootstrap servers we consider a disabled kafka extension
+			kafka.ExtDisabled = constants.KIE_KAFKA_EXT_DISABLED_DEFAULT
+			kafka.BootstrapServers = constants.KIE_KAFKA_EXT_BOOTSTRAP_SERVERS_DEFAULT
+		}
+
+		if kafka.MaxBlockMs == 0 {
+			kafka.MaxBlockMs = constants.KIE_KAFKA_EXT_MAX_BLOCK_MS_DEFAULT
+		}
+
+		if kafka.GroupID == "" {
```
spolti, line 1315: for strings it is better to use len(xxx) == 0 or < 0..

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1293,6 +1299,28 @@ func setJvmDefault(jvm *api.JvmObject) {
 	}
 }
 
+func setKafkaExtDefault(kafka *api.KafkaExt) {
+	if kafka != nil {
+
+		if kafka.BootstrapServers == "" {
```
spolti, line 1305: for strings it is better to use len(xxx) == 0 or < 0..

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1293,6 +1299,28 @@ func setJvmDefault(jvm *api.JvmObject) {
...
+			//if aren't present bootstrap servers we consider a disabled kafka extension
+			kafka.ExtDisabled = constants.KIE_KAFKA_EXT_DISABLED_DEFAULT
```
spolti, line 1307: there is a reason to set it here if, at this point, if kafka object exists, this env should be true already?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1307,6 +1335,13 @@ func checkJvmOnServer(server *api.KieServerSet) {
 	setJvmDefault(server.Jvm)
 }
 
+func checkKafkaExt(server *api.KieServerSet) {
+	if server.KafkaExt == nil {
```
spolti, line 1339: by doing this, the kafka object will always be set, it is something expected?
I think the JMS integration only gets executed if if the expected property is enabled, might be interesting apply something similar for kafka.

config/7.11.0/common.yaml
```diff
@@ -829,6 +829,26 @@ servers:
                       value: "[[.Jvm.GcContainerOptions]]"
                     #[[end]]
                     ## Jvm config END
+                    ## AMQ Streams BEGIN
+                    - name: KIE_SERVER_KAFKA_EXT_DISABLED
+                      value: "[[.KafkaExt.ExtDisabled]]"
+                    #[[if not .KafkaExt.ExtDisabled]]
```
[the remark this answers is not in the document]
spolti: please make sure that all other properties are unset, if empty the envs shouldn't be set.

spolti (changes requested):

pkg/controller/kieapp/constants/constants.go
desmax74: Yep, only test, moving there

pkg/controller/kieapp/constants/constants.go
desmax74: Ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: yep, was a wrong consideration about UI, but could be removed

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -942,6 +944,26 @@ type ObjectReference struct {
 	FieldPath string `json:"fieldPath,omitempty" protobuf:"bytes,7,opt,name=fieldPath"`
 }
 
+// KafkaExt kafka configuration to be used by the KieApp
+type KafkaExt struct {
+	// Disable the Kafka Extension
+	ExtDisabled bool `json:"extDisabled,omitempty"`
+	// A comma separated list of pairs <message/signal name>=<topicName>
+	Topics []corev1.EnvVar `json:"topics,omitempty"`
```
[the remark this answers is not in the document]
desmax74: In the UI with kieserver button was used an array of Env, I'll check if it works with strings

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -942,6 +944,26 @@ type ObjectReference struct {
...
+	// A comma separated list of bootstrapServers
+	BootstrapServers string `json:"bootstrapServers,omitempty"`
```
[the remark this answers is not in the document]
desmax74: Here not, because is psased to kafka a string like this "localhost:9092,another.host:9092"

deploy/ui/form.json
```diff
@@ -848,6 +848,97 @@
...
                 },
+                {
+                  "label": "Kafka Extension",
+                  "required": false,
+                  "jsonPath": "$.spec.objects.servers[*].kafkaExt",
+                  "type": "object",
+                  "max": 1,
+                  "fields": [
+                    {
+                      "label": "Kafka Extension",
```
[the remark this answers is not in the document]
desmax74: Unfortunately not, we haven't a full control over UI, we can't set a default value in a input text or have conditional behaviourm and the console-form-cr will be deprecated asap to use the openshift ui only

deploy/ui/form.json
```diff
@@ -848,6 +848,97 @@
...
+                      "type": "checkbox",
+                      "checked": "true",
+                      "required": false,
+                      "jsonPath": "$.spec.objects.servers[*].kafkaExt.extDisabled",
+                      "originalJsonPath": "$.spec.objects.servers[*].kafkaExt.extDisabled",
+                      "description": "Disable Kafka extension."
+                    },
+                    {
+                      "label": "BootstrapServers",
```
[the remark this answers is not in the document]
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok, but  == "" or != "" seems used https://github.com/kiegroup/kie-cloud-operator/blob/main/pkg/controller/kieapp/defaults/defaults.go#L966

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

config/7.11.0/common.yaml
```diff
@@ -829,6 +829,26 @@ servers:
                       value: "[[.Jvm.GcContainerOptions]]"
                     #[[end]]
                     ## Jvm config END
+                    ## AMQ Streams BEGIN
+                    #[[if .Kafka.ExtEnabled]]
+                    - name: KIE_SERVER_KAFKA_EXT_ENABLED
+                      value: "[[.Kafka.ExtEnabled]]"
+                    - name: KIE_SERVER_KAFKA_EXT_TOPICS
+                      value: "[[.Kafka.Topics]]"
```
tchughesiv, line 837: if the bulk of the kafka fields are optional, it feels like we should have [[if]] statements for each one of these options in the common.yaml. Otherwise, the operator will insert blank values for each one. What would happen to the app if that happened? if these are all required, then that should be changed in the crd.

pkg/apis/app/v2/kieapp_types.go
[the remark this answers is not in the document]
tchughesiv: you can pass to kafka that way via code though... and still ask for this as a string list in kieapp cr. just a thought. don't really have an opinion on which way is better.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1010,6 +1018,28 @@ func getJmsConfig(environment api.EnvironmentType, jms *api.KieAppJmsObject) (*a
 	return jms, nil
 }
 
+func getKafkaConfig(environment api.EnvironmentType, kafka *api.KafkaExtObject) (*api.KafkaExtObject, error) {
+	envConstants := constants.EnvironmentConstants[environment]
```
tchughesiv, line 1022: any reason you're pulling the envConstants here? doesn't seem you've added any kafka specific items to that map, so maybe not needed?

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
desmax74: yep, remove

config/7.11.0/common.yaml
desmax74: The envs depends on .Kafka.ExtEnabled is is enabled the env will be passed, we are considering all or nothing, depending on the extEnabled

tchughesiv: so all are required then?

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5434,3 +5434,129 @@ func TestClusterLabelsRHPAMDashbuilderDefaultEnvironment(t *testing.T) {
...
+
+	for _, env := range envs {
+
+		if env.Name == "KIE_SERVER_KAFKA_EXT_ENABLED" {
+			extEnabled = true
+			assert.True(t, env.Value == "false")
+		}
+	}
+	assert.True(t, extEnabled)
+}
```
tchughesiv, line 5562: we need to add a check somewhere for what happens when the kafka object is empty. the object should be empty in spec and status, and the env vars should not exist.

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5434,3 +5434,147 @@ func TestClusterLabelsRHPAMDashbuilderDefaultEnvironment(t *testing.T) {
...
+	cr := &api.KieApp{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test",
+		},
+		Spec: api.KieAppSpec{
+			Environment: api.RhdmProductionImmutable,
+			Objects: api.KieAppObjects{
+				Servers: []api.KieServerSet{
+					{
+						Kafka: &api.KafkaExtObject{},
```
tchughesiv, line 5541: feels like the better test would be to remove this line and just check that kafka in status is empty

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: ok

pkg/controller/kieapp/constants/constants.go
```diff
@@ -260,6 +260,16 @@ const (
 	//Envs
 	KubeNS     = "KUBERNETES_NAMESPACE"
 	KubeLabels = "KUBERNETES_LABELS"
+
+	//Kafka extension
+	KieServerKafkaExtEnabled          = "KIE_SERVER_KAFKA_EXT_ENABLED"
```
spolti, line 265: besides tests where else is this constants being used?

spolti (changes requested):

pkg/controller/kieapp/constants/constants.go
desmax74: Just for test, moving in the tests

deploy/crds/kieapp.crd.yaml
```diff
@@ -4018,6 +4058,46 @@ spec:
                                     to generated options in JAVA_OPTS. e.g. '-Dsome.property=foo'
                                   type: string
                               type: object
+                            kafka:
```
spolti, line 4061: update the field descriptions to match the added on the templates.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1010,6 +1018,27 @@ func getJmsConfig(environment api.EnvironmentType, jms *api.KieAppJmsObject) (*a
...
+		return nil, nil
+	}
+
+	defaultKafka := api.KafkaExtObject{
+		MaxBlockMs:       200,
+		AutocreateTopics: Pbool(true),
+		BootstrapServers: "localhost:9092",
+		GroupID:          "jbpm-consumer",
+		Acks:             2,
+	}
```
tchughesiv, line 1032: are we not allowing the user to set everything via the cr spec? it appears you're setting defaults and its not configurable?

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
desmax74: Removed

deploy/crds/kieapp.crd.yaml
desmax74: done

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: i didn't mean to remove it... instead there should be logic that checks if the field is blank before setting a default. if its not blank, then the current setting should be used. right?

desmax74: Added, we set missing values and override if something has changed

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1010,6 +1019,44 @@ func getJmsConfig(environment api.EnvironmentType, jms *api.KieAppJmsObject) (*a
...
+
+	var override = false
+	defaultKafka := api.KafkaExtObject{}
+	//if something is missing we set mandatory defaults
+	if kafka.MaxBlockMs == 0 {
+		defaultKafka.MaxBlockMs = 2000
+		override = true
+	}
+	if len(kafka.BootstrapServers) == 0 {
+		defaultKafka.BootstrapServers = "localhost:9092"
```
spolti, line 1035: I don't think it a good idea to set a default value for the server address, probably better make this field mandatory and fail the kafka setup if it is not set.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1534,6 +1581,34 @@ func mergeDashbuilder(service kubernetes.PlatformService, cr *api.KieApp, env ap
 	return env, nil
 }
 
+func overrideKafkaTopicsEnv(cr *api.KieApp, env *api.Environment) {
+	var dataSet []string
+	if cr.Status.Applied.Objects.Servers[0].Kafka != nil {
+		if cr.Status.Applied.Objects.Servers != nil {
```
spolti, line 1587: shouldn't this if condition be before? than kafka struct test?

spolti (changes requested):

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1010,6 +1019,44 @@ func getJmsConfig(environment api.EnvironmentType, jms *api.KieAppJmsObject) (*a
...
+	if kafka == nil || !kafka.ExtEnabled {
+		return nil, nil
+	}
+
+	var override = false
+	defaultKafka := api.KafkaExtObject{}
+	//if something is missing we set mandatory defaults
+	if kafka.MaxBlockMs == 0 {
+		defaultKafka.MaxBlockMs = 2000
+		override = true
```
tchughesiv, line 1032: instead of using defaultKafka and override, why not just set `kafka.MaxBlockMs = 2000` directly? you could do the same for the rest of the defaults and then get rid of the mergo logic. is there something i'm missing?

tchughesiv (changes requested):

config/7.11.0/common.yaml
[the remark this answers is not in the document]
tchughesiv: @spolti is this still an ask? I assume you want the same type of if statement added for each env var?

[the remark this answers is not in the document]
spolti: Yes.
I probably was not clear in my first comment in this regard, I've suggested just move the condition to test if the kafka integration is enabled to the top in a way if disable the other fields wouldn't be checked and if enabled, each field would have its own if condition to only set the env if the value is present.

pkg/controller/kieapp/defaults/defaults.go
desmax74: Ok I was thinking that mergo was the pattern to set defaults

pkg/controller/kieapp/defaults/defaults.go
desmax74: yes

pkg/controller/kieapp/defaults/defaults.go
desmax74: Is asked here as default value https://issues.redhat.com/browse/RHPAM-3331?focusedCommentId=15539246&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15539246

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1015,6 +1024,31 @@ func getJmsConfig(environment api.EnvironmentType, jms *api.KieAppJmsObject) (*a
 	return jms, nil
 }
 
+func getKafkaConfig(kafka *api.KafkaExtObject) (*api.KafkaExtObject, error) {
+	if kafka == nil || !kafka.ExtEnabled {
+		return nil, nil
+	}
+
+	//if something is missing we set mandatory defaults
+	if kafka.MaxBlockMs == 0 {
```
tchughesiv, line 1033: would making this a pointer and checking for nil be a better way to go about this?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1015,6 +1024,31 @@ func getJmsConfig(environment api.EnvironmentType, jms *api.KieAppJmsObject) (*a
...
+		kafka.MaxBlockMs = 2000
+	}
+	if len(kafka.BootstrapServers) == 0 {
+		kafka.BootstrapServers = "localhost:9092"
+	}
+	if len(kafka.GroupID) == 0 {
+		kafka.GroupID = "jbpm-consumer"
+	}
+	if kafka.Acks == 0 {
```
tchughesiv, line 1042: same

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1539,6 +1573,32 @@ func mergeDashbuilder(service kubernetes.PlatformService, cr *api.KieApp, env ap
 	return env, nil
 }
 
+func overrideKafkaTopicsEnv(cr *api.KieApp, env *api.Environment) {
+	var dataSet []string
+	if cr.Status.Applied.Objects.Servers != nil && cr.Status.Applied.Objects.Servers[0].Kafka != nil {
```
tchughesiv, line 1578: checking `cr.Status.Applied.Objects.Servers[0].Kafka` doesn't work here because what if `cr.Status.Applied.Objects.Servers[1].Kafka` is not nil instead? each serverSet should have the ability to set enable Kafka if it wants.

tchughesiv (changes requested):

spolti:
hold until https://github.com/jboss-container-images/jboss-kie-modules/pull/446 is not merged.

pkg/controller/kieapp/defaults/defaults.go
desmax74: Yep

pkg/controller/kieapp/defaults/defaults.go
desmax74: yep

pkg/controller/kieapp/defaults/defaults.go
desmax74: Ok

config/7.11.0/common.yaml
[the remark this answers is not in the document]
tchughesiv: @desmax74 fyi... this needs to happen as well.

spolti:
@desmax74 please rebase to solve the conflict.

config/7.11.0/common.yaml
spolti: @desmax74 any update on this?

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5520,3 +5520,164 @@ func TestRhdmProdImmutableEnvironmentWithoutJbpmCluster(t *testing.T) {
...
+			},
+		},
+	}
+
+	env, err := GetEnvironment(cr, test.MockService())
+	assert.Nil(t, err, "Error getting prod environment")
+	assert.NotNil(t, env)
+	assert.True(t, len(cr.Spec.Objects.Servers[0].Kafka.Topics) == 2)
+	envs := env.Servers[0].DeploymentConfigs[0].Spec.Template.Spec.Containers[0].Env
+	autocreateEnv, maxBlockMS, topics, extEnabled, clientID, acks, groupID, bootstrapServers := false, false, false, false, false, false, false, false
```
spolti, line 5546: what is the purpose of this line?

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5520,3 +5520,164 @@ func TestRhdmProdImmutableEnvironmentWithoutJbpmCluster(t *testing.T) {
...
+		case "KIE_SERVER_KAFKA_EXT_SERVERS":
+			bootstrapServers = true
+			assert.True(t, env.Value == "localhost:9092")
+
+		case "KIE_SERVER_KAFKA_EXT_TOPICS":
+			topics = true
+			assert.True(t, env.Value == "events=my-topics,errors=my-errs")
+		}
+	}
+	assert.True(t, autocreateEnv)
```
spolti, line 5582: not sure if I understood the purpose of these assertions here.

pkg/controller/kieapp/defaults/defaults_test.go
spolti: nil or empty?

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5520,3 +5520,164 @@ func TestRhdmProdImmutableEnvironmentWithoutJbpmCluster(t *testing.T) {
...
+			},
+		},
+	}
+
+	env, err := GetEnvironment(cr, test.MockService())
+	assert.Nil(t, err, "Error getting environment")
+	assert.NotNil(t, env)
+	envs := env.Servers[0].DeploymentConfigs[0].Spec.Template.Spec.Containers[0].Env
+	extEnabled := false
+	for _, env := range envs {
```
spolti, line 5675: @tchughesiv IIRC there are auxiliary functions to return the env value, right?

spolti (changes requested):

deploy/ui/form.json
```diff
@@ -1162,6 +1162,89 @@
...
                 },
+                {
+                  "label": "Enable Kafka Extension",
+                  "type": "object",
+                  "max": 1,
+                  "jsonPath": "$.spec.objects.servers[*].kafka",
+                  "visible": true,
+                  "fields": [
+                    {
+                      "label": "Kakfa extension",
```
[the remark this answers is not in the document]
desmax74: Fixed

deploy/ui/form.json
```diff
@@ -848,6 +848,97 @@
                   "type": "integer",
                   "description": "The number of replicas for this KIE Server set."
                 },
+                {
+                  "label": "Kafka Extension",
```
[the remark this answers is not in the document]
desmax74: Done

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1539,6 +1573,32 @@ func mergeDashbuilder(service kubernetes.PlatformService, cr *api.KieApp, env ap
 	return env, nil
 }
 
+func overrideKafkaTopicsEnv(cr *api.KieApp, env *api.Environment) {
+	var dataSet []string
```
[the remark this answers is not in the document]
desmax74: renamed

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: Checks if all envs are present with the expected values

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: Checks if all envs are present with the expected values

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: In htis test empty and in another is nil

config/7.11.0/common.yaml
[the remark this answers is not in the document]
desmax74: @spolti @tchughesiv done

config/7.11.0/common.yaml
desmax74: @spolti is it ok in this way ?

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

deploy/olm-catalog/dev/7.11.0-1/manifests/kieapp.crd.yaml
```diff
@@ -1876,6 +1876,47 @@ spec:
                                 to generated options in JAVA_OPTS. e.g. '-Dsome.property=foo'
                               type: string
                           type: object
+                        kafka:
```
desmax74, line 1879: @tchughesiv  this addition was added during the build after the rebase is it ok to add ?

deploy/olm-catalog/prod/7.11.0-1/manifests/kieapp.crd.yaml
```diff
@@ -1876,6 +1876,47 @@ spec:
                                 to generated options in JAVA_OPTS. e.g. '-Dsome.property=foo'
                               type: string
                           type: object
+                        kafka:
```
desmax74, line 1879: @tchughesiv  this addition was added during the build after the rebase is it ok to add ?

deploy/olm-catalog/test/7.11.0-1/manifests/kieapp.crd.yaml
```diff
@@ -1876,6 +1876,47 @@ spec:
                                 to generated options in JAVA_OPTS. e.g. '-Dsome.property=foo'
                               type: string
                           type: object
+                        kafka:
```
desmax74, line 1879: @tchughesiv  this addition was added during the build after the rebase is it ok to add ?

deploy/olm-catalog/dev/7.11.0-1/manifests/kieapp.crd.yaml
spolti: it is a reflection of what you have added to kieapp_types.
It is expected.

deploy/ui/form.json
```diff
@@ -1180,6 +1180,89 @@
...
                 },
+                {
+                  "label": "Enable Kafka Extension",
+                  "type": "object",
+                  "max": 1,
+                  "jsonPath": "$.spec.objects.servers[*].kafka",
+                  "visible": true,
+                  "fields": [
+                    {
+                      "label": "Enable Kafka extension",
```
spolti, line 1191: @tchughesiv is there a way to make a field invisible?
this one would make sense as the button before explicit "enable" this feature.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -141,6 +141,7 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 	if err != nil {
 		return api.Environment{}, err
 	}
+	overrideKafkaTopicsEnv(cr, &mergedEnv)
```
spolti, line 144: I think it should be inside getKafkaConfig. isn't it better/;

pkg/controller/kieapp/defaults/defaults_test.go
spolti: I think there is a better way to test envs, see: 
https://github.com/kiegroup/kie-cloud-operator/blob/main/pkg/controller/kieapp/defaults/defaults_test.go#L406

pkg/controller/kieapp/defaults/defaults_test.go
spolti: see my comment about, there is a better way to do this.

spolti (changes requested):

deploy/ui/form.json
tchughesiv: it would seem so... see line #1188 above...
`"visible": false,`?

tchughesiv: i do wonder though... is the `extEnabled` field even necessary given the kafa object is a pointer? couldn't we just use that pointer as a means to determine whether to enable? if a user is setting anything in the kafka object, that would mean they want it enabled right?

pkg/controller/kieapp/defaults/defaults.go
tchughesiv: this needs to be explored more

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5673,3 +5673,164 @@ func TestRhdmProdImmutableEnvironmentWithoutJbpmCluster(t *testing.T) {
...
+
+	extEnabled := false
+	for _, env := range envs {
+		if strings.HasPrefix(env.Value, "KIE_SERVER_KAFKA") {
+			extEnabled = true
+		}
+	}
+	assert.False(t, extEnabled)
+
+	kafkaSpec := cr.Spec.Objects.Servers[0].Kafka
```
tchughesiv, line 5789: this should be `kafkaSpec := cr.Status.Applied.Objects.Servers[0].Kafka`

we care about what's being applied. we already know what was set in the spec.

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: Ok

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: yep

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: IS present here: https://github.com/kiegroup/kie-cloud-operator/pull/614/files/6688118e35fe8ba23f66aeb7eb05c6ce10b6b2ba#diff-20d788a31791c7601d275bf50c8e30cb790834f4c05ce64c4e3bdfe92a6aaebaR5799

deploy/ui/form.json
desmax74: the visible is on a fieldGroup i.e. the button to enable something, not on a single field like checkbox or text

tchughesiv:
@desmax74 fyi
https://github.com/desmax74/kie-cloud-operator/pull/6

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5673,3 +5673,157 @@ func TestRhdmProdImmutableEnvironmentWithoutJbpmCluster(t *testing.T) {
...
+			assert.True(t, env.Value == "2100")
+
+		case "KIE_SERVER_KAFKA_EXT_CLIENT_ID":
+			assert.True(t, env.Value == "C1234567")
+
+		case "KIE_SERVER_KAFKA_EXT_SERVERS":
+			assert.True(t, env.Value == "localhost:9092")
+
+		case "KIE_SERVER_KAFKA_EXT_TOPICS":
+			assert.True(t, env.Value == "events=my-topics,errors=my-errs")
```
tchughesiv, line 5725: switch these all to `assert.Equal`... it's more clear when the test fails.

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5673,3 +5673,157 @@ func TestRhdmProdImmutableEnvironmentWithoutJbpmCluster(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	env, err := GetEnvironment(cr, test.MockService())
+	assert.Nil(t, err, "Error getting prod environment")
+	assert.NotNil(t, env)
+	assert.True(t, len(cr.Spec.Objects.Servers[0].Kafka.Topics) == 2)
```
tchughesiv, line 5697: assert.Len

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -5673,3 +5673,157 @@ func TestRhdmProdImmutableEnvironmentWithoutJbpmCluster(t *testing.T) {
...
+
+func TestRhdmEnvironmentWithKafkaExtDefault(t *testing.T) {
+	cr := &api.KieApp{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test",
+		},
+		Spec: api.KieAppSpec{
+			Environment: api.RhdmProductionImmutable,
+			Objects: api.KieAppObjects{
+				Servers: []api.KieServerSet{
```
tchughesiv, line 5760: we need to verify what happens w/ multiple servers configured... maybe set one w/ servers[0].name so you can track the configs better in status. the same kafka configs should stay put in status.

tchughesiv (changes requested):

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: Three servers, one with kafka default, one without kafka and one with kafka configured

desmax74: Three server, one with default, one without kafka and one configured, could be ok ?

deploy/ui/form.json
```diff
@@ -1180,6 +1180,80 @@
...
+                      "required": false,
+                      "jsonPath": "$.spec.objects.servers[*].kafka.maxBlockMs",
+                      "originalJsonPath": "$.spec.objects.servers[*].kafka.maxBlockMs",
+                      "description": "Number of milliseconds that indicates how long publish method will block. Default 2000 (2 seconds)"
+                    },
+                    {
+                      "label": "Topics",
+                      "required": false,
+                      "jsonPath": "$.spec.objects.servers[*].kafka.topics",
+                      "type": "object",
```
tchughesiv, line 1243: is this type right now that its a string slice?

@swati-kale what type should be used here for a string array/slice? []string?

tchughesiv (changes requested):

config/7.11.0/common.yaml
spolti: yes.

deploy/ui/form.json
tchughesiv: is an invisible field this still necessary now that we're using a pointer for template?

deploy/ui/form.json
swati-kale: Console CR form supports array of objects, where objects may have any fields e.g string.

swati-kale: if you want the yaml representation as  below for topics ,
topics:
            - xyz
            - abc
then I will suggest changes

swati-kale: "fields": [
                        {
                          "label": "Mapping Message=Topic",
                          "type": "text",
                          "required": false,
                          "jsonPath": "$.spec.objects.servers[*].kafka.topics[*]",
                          "originalJsonPath": "$.spec.objects.servers[*].kafka.topics[*]",
                          "description": "Mapping Message=Topic"
                        }

tchughesiv (approved):

PR #647, opened 2021-05-03
[RHPAM-3640] Configure Kafka JBPM Emitter from Operator

desmax74:
7.11.0 CRX
See: https://issues.redhat.com/browse/RHPAM-3640

@spolti @jakubschwan


deploy/crds/kieapp.crd.yaml
```diff
@@ -1913,6 +1913,31 @@ spec:
                                 type: string
                               type: array
                           type: object
+                        kafkaJbpm:
```
jakubschwan, line 1916: To make configuration cleaner, I would suggest to use the whole name. kafkaJbpm mayn not be clear and in some cases confusing, e.g. users would set up kafkaJbpm and expects that the messages from process would be handled.
```suggestion
                        kafkaJbpmEventEmitters:
```

deploy/crds/kieapp.crd.yaml
```diff
@@ -1913,6 +1913,31 @@ spec:
...
+                              description: Jbpm Emitter, identifier to pass to the
+                                server when making requests
+                              type: string
+                            maxBlockMs:
+                              description: Jbpm emitter, number of milliseconds that
+                                indicates how long publish method will bloc for jbpm
+                                emitter
+                              format: int32
+                              type: integer
+                          type: object
```
jakubschwan, line 1940: what about configuration of these system properties via Operator:

- date format of message: org.kie.jbpm.event.emitters.kafka.date_format: The timestamp format for the time field of the messages. The default value is yyyy-MM-dd’T’HH:mm:ss.SSSZ 
- mapping of topic name for processes org.kie.jbpm.event.emitters.kafka.topic.processes: The topic name for process event messages. The default value is jbpm-processes-events.
- mapping of topic name for casesorg.kie.jbpm.event.emitters.kafka.topic.cases: The topic name for process event messages. The default value is jbpm-cases-events.
- mapping of topic name for tasks org.kie.jbpm.event.emitters.kafka.topic.tasks: The topic name for process event messages. The default value is jbpm-processes-tasks.

deploy/ui/form.json
```diff
@@ -1254,6 +1254,46 @@
                     }
                   ]
                 },
+                {"label": "Enable Kafka Emitter",
```
jakubschwan, line 1257: formatting change
```suggestion
                {
                  "label": "Enable Kafka Emitter",
```

config/7.11.0/common.yaml
```diff
@@ -940,6 +940,24 @@ servers:
                       value: "[[.Kafka.MaxBlockMs]]"
                       #[[end]]
                     #[[end]]
+                    #[[if .KafkaJbpm]]
```
jakubschwan, line 943: It would be good to have there also some flag, like we have for Kafak integratino KIE_SERVER_KAFKA_EXT_ENABLED. We would need custom flag for KafkaJbpmEventEmitter (e.g. `KIE_SERVER_KAFKA_JBPM_EVENT_EMITTER_ENABLED`) so then we could easily check in Kie Server image that a new configuration is needed and emitter jar needs to be moved to the correct place in ROOT.war

config/7.11.0/common.yaml
```diff
@@ -940,6 +940,24 @@ servers:
...
+                    - name: KIE_SERVER_KAFKA_JBPM_ACKS
+                      value: "[[.KafkaJbpm.Acks]]"
+                      #[[end]]
+                      #[[if .KafkaJbpm.ClientID]]
+                    - name: KIE_SERVER_KAFKA_JBPM_CLIENT_ID
+                      value: "[[.KafkaJbpm.ClientID]]"
+                      #[[end]]
+                      #[[if .KafkaJbpm.MaxBlockMs]]
+                    - name: KIE_SERVER_KAFKA_JBPM_MAX_BLOCK_MS
+                      value: "[[.KafkaJbpm.MaxBlockMs]]"
```
jakubschwan, line 958: same as for the KieApp objects, I think that it would be good to use full names also for Kie Server env. variables. So instead of KAFKA_JBPM to use full naming -> KAFKA_JBPM_EVENT_EMITTER

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1073,6 +1077,19 @@ func getKafkaConfig(kafka *api.KafkaExtObject) {
 	}
 }
 
+func getKafkaJBPMConfig(kafkaExt *api.KafkaExtObject, kafka *api.KafkaJBPMObject) {
+	// The emitter is enabled only if the kafka extension is enabled
+	if kafkaExt != nil && kafka != nil {
```
jakubschwan, line 1082: please check this with guys from jbpm, the configuration through system properties looks like that it may be possible to use only kafka jbpm event emitter

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1073,6 +1077,19 @@ func getKafkaConfig(kafka *api.KafkaExtObject) {
...
+		//if something is missing we set mandatory defaults
```
jakubschwan, line 1083: thinking if this is really needed, because system properties already have predefined default values... on the other hand users can see in the KieApp (in applied) what exactly is set in their env. Our defaults are same as defaults set in system properties - see [docs](https://access.redhat.com/documentation/en-us/red_hat_process_automation_manager/7.10/html-single/integrating_red_hat_process_automation_manager_with_other_products_and_components/index#kieserver-kafka-emit-proc_integrating-amq-streams)
@spolti wdyt?

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1073,6 +1077,19 @@ func getKafkaConfig(kafka *api.KafkaExtObject) {
...
+		if kafka.MaxBlockMs == nil {
+			kafka.MaxBlockMs = Pint32(2000)
+		}
+		if len(kafka.BootstrapServers) == 0 {
+			kafka.BootstrapServers = "localhost:9092"
+		}
```
jakubschwan, line 1089: some of the KafkaJbpmEventEmitter properties are missing, if we are providing defaults, we should add there all of them (same as we have for Kafka extension)

jakubschwan (changes requested):

config/7.11.0/common.yaml
desmax74: We could set but in the jboss-kie-modules because in the operator if the user open the kafka extension we are considering enabled (if configured correctly) see https://github.com/jboss-container-images/jboss-kie-modules/blob/master/jboss-kie-kieserver/added/launch/jboss-kie-kieserver.sh#L760

config/7.11.0/common.yaml
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: we are using default sonly on the properties we know the defaults, bootstraps and acks

deploy/ui/form.json
desmax74: ok

deploy/crds/kieapp.crd.yaml
desmax74: - date format of message: value of date format is readed from the env "org.kie.server.json.date_format" https://github.com/kiegroup/jbpm/blob/master/jbpm-event-emitters/jbpm-event-emitters-kafka/src/main/java/org/jbpm/event/emitters/kafka/KafkaEventEmitter.java#L71 isn't a good idea override
- mapping of topic name for processes/case/tasks are fixed https://github.com/kiegroup/jbpm/blob/master/jbpm-event-emitters/jbpm-event-emitters-kafka/src/main/java/org/jbpm/event/emitters/kafka/KafkaEventEmitter.java#L163 we could map in the jboss-kie-modules

config/7.11.0/common.yaml
desmax74: ok

deploy/crds/kieapp.crd.yaml
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: Asked to Francisco, waiting a reply

desmax74: Asked to Francisco, are separate things

spolti:
/hold

waiting for the decision about what version these changes will be added.

pkg/controller/kieapp/defaults/defaults.go
spolti: if kieserver is setting defaults already, I agree to not set them here, just add in the field description what are the default values if empty.

pkg/controller/kieapp/defaults/defaults.go
spolti: I don't like this default value for bootstrap servers, for ocp, it will never work.
better make the property mandatory and fail the reconcile loop if it is not set.

spolti (changes requested):

deploy/crds/kieapp.crd.yaml
```diff
@@ -1913,6 +1913,31 @@ spec:
...
+                                connection to the Kafka cluster for Jbpm Emitter
+                              type: string
+                            clientID:
+                              description: Jbpm Emitter, identifier to pass to the
+                                server when making requests
+                              type: string
+                            maxBlockMs:
+                              description: Jbpm emitter, number of milliseconds that
+                                indicates how long publish method will bloc for jbpm
+                                emitter
```
jakubschwan, line 1937: Missing information about default value

deploy/crds/kieapp.crd.yaml
```diff
@@ -1913,6 +1913,31 @@ spec:
...
                               type: array
                           type: object
+                        kafkaJbpmEventEmitters:
+                          description: KafkaJBPMEventEmittersObject kafka configuration
+                            to be used by the KieApp for JBPM Emitter
+                          properties:
+                            acks:
+                              description: Jbpm Emitter, the number of acknowledgments
+                                the producer requires the leader to have received
+                                before considering a request complete.
```
jakubschwan, line 1923: Missing information about default value

deploy/crds/kieapp.crd.yaml
```diff
@@ -1913,6 +1913,31 @@ spec:
...
+                              type: integer
+                            bootstrapServers:
+                              description: Jbpm emitter, a comma separated list of
+                                host/port pairs to use for establishing the initial
+                                connection to the Kafka cluster for Jbpm Emitter
+                              type: string
+                            clientID:
+                              description: Jbpm Emitter, identifier to pass to the
+                                server when making requests
```
jakubschwan, line 1932:
```suggestion
                                server when making requests.
```

deploy/crds/kieapp.crd.yaml
jakubschwan, line 1937:
```suggestion
                                emitter.
```

deploy/crds/kieapp.crd.yaml
desmax74: Is autogenerated the crd file

deploy/crds/kieapp.crd.yaml
desmax74: Is autogenerated the crd file

deploy/crds/kieapp.crd.yaml
desmax74: Is autogenerated the crd file

PR #659, opened 2021-06-22
[RHPAM-3763] - Endless updating loop in Kie operator

spolti:

### DIFF
--- a/hack/go-build-bundle.sh
+++ b/hack/go-build-bundle.sh
@@ -19,15 +19,18 @@ if [[ ${LOCAL} != true ]]; then
     fi
 fi
 
-echo ${CFLAGS}
+echo "Cekit build flags : ${CFLAGS}"
 
 OLMDIR=deploy/olm-catalog/prod
 CSV=businessautomation-operator.clusterserviceversion.yaml
 if [[ ${DEV} == true ]]; then
     OLMDIR=deploy/olm-catalog/dev
-    BUNDLE_NAME=quay.io/tchughesiv/${BUNDLE}
+    BUNDLE_NAME=quay.io/${USERNAME}/${BUNDLE}
 fi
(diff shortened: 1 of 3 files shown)


hack/go-build-bundle.sh
```diff
@@ -78,7 +81,7 @@ if [[ ${LOCAL} != true ]]; then
 else
     cekit -v --descriptor image-bundle.yaml --redhat build \
         --overrides '{name: '${BUNDLE_NAME}'}' \
-        --overrides '{version: '${VERSION}'}' \
+        --overrides '{version: '${CSVVERSION}'}' \
```
tchughesiv, line 84: why changing this? image tag needs to be `7.11.0`, not 7.11.0-2... right?

hack/go-build-bundle.sh
```diff
@@ -54,7 +57,7 @@ cekit-cache add --md5 ${MD5_ANNO} ${ANNO_PATH}
 if [[ ${LOCAL} != true ]]; then
     cekit -v --descriptor image-bundle.yaml --redhat build \
         --overrides '{name: '${BUNDLE_NAME}'}' \
-        --overrides '{version: '${VERSION}'}' \
+        --overrides '{version: '${CSVVERSION}'}' \
```
tchughesiv, line 60: same

tchughesiv (changes requested):

hack/go-build-bundle.sh
spolti: I have updated because of this the instructions on the README points the bundle image tag to 7.xx.y-z

Btw, before merging this I should increase the tag version, right?

tchughesiv (changes requested):

hack/go-build-bundle.sh
spolti: I see, will revert it back, thanks.

PR #663, opened 2021-06-23
remove previous tag files

spolti:


spolti:
hey @tchughesiv, this is the same scenario you mentioned to me that the config test would fail, only when removing previous config versions, right? In this case, is this one good to go?

tchughesiv:
/lgtm
@spolti correct

PR #634, opened 2021-04-09
[RHPAM-3279] Configure CORS automatically through a flag or environment variable

desmax74:
Signed-off-by: desmax74 <mdessi@redhat.com>
See: https://issues.redhat.com/browse/RHPAM-3279

@spolti @tchughesiv @jakubschwan

PR #669, opened 2021-07-13
[RHPAM-3279] Configure CORS automatically through a flag or environment 

desmax74:
See: https://issues.redhat.com/browse/RHPAM-3279


spolti (changes requested):
I am not sure if I agree with the implementation done here. The idea behind the feature request would be to have it more dynamic. Basically, based on this implementation user can only enable or disable (https://github.com/kiegroup/kie-cloud-operator/pull/669/files#diff-ba2395bb079300b28d21e2de1bd4999a4636be3ba314daac32b1962caf571448R982) it and default values will set for all filters. If I want, for example, enable only the AC_ALLOW_ORIGIN filter, that would not be possible. And the UI does not have any of the fields available. 

Another point, if I, for example, add in the kieApp CR `cors.Filters` to have only the AC_ALLOW_HEADERS all other envs to other filters will be added: https://github.com/kiegroup/kie-cloud-operator/pull/669/files#diff-ba2395bb079300b28d21e2de1bd4999a4636be3ba314daac32b1962caf571448R983-R1004

Please rework this PR to achieve this.

Other thing, Business/Decision Central and Business Central moniitoring also needs to have this configuration, not only KIE Server.

desmax74:
@spolti 
We have choosed this PR with the checkbox to enable the CORS because the other PR with all the field can not display the default values in each checkbox https://github.com/kiegroup/kie-cloud-operator/pull/634/ due to the UI limitation.

The CORS filters works together isn't possible add some filters and not others, or your idea is to overwrite the input edited by the user ans set the others with default value ?

desmax74:
Replaced by https://github.com/kiegroup/kie-cloud-operator/pull/634

PR #671, opened 2021-07-26
[RHPAM-3591] - Persist KIE Server maven and kie repositories

spolti:

See: https://issues.redhat.com/browse/RHPAM-3591

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -704,6 +704,9 @@ func getServersConfig(cr *api.KieApp) ([]api.ServerTemplate, error) {
 				KeystoreSecret:   serverSet.KeystoreSecret,
 				StorageClassName: serverSet.StorageClassName,
 				JbpmCluster:      serverSet.JbpmCluster,
+				PersistRepos:     serverSet.PersistRepos,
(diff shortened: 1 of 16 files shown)


msmagnanijr (approved):
lgtm

desmax74 (approved):

desmax74:
lgtm

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1701,6 +1716,14 @@ func isImmutable(cr *api.KieApp) bool {
 	return false
 }
 
+func isTrial(cr *api.KieApp) bool {
```
vaibhavjainwiz, line 1719: What is this trial mode?

vaibhavjainwiz (commented):
Just one doubt. All looks good apart from it.

pkg/controller/kieapp/defaults/defaults.go
spolti: The Trial mode is the most simplier combination we can have when deploying RHPAM/DM.
It deploys the KIE Server and the Business/Decision Central with no PV and database used is embedded, the H2.
plus, it uses http routes.

vaibhavjainwiz: thanks for explaining it.

PR #731, opened 2022-01-20
[RHPAM-3787] - DockerImage kind is not accepted in Business Automatio…

spolti:
…n operator with JDBC extension images

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults_test.go
+++ b/pkg/controller/kieapp/defaults/defaults_test.go
@@ -3293,6 +3293,14 @@ func TestSetKieServerFrom(t *testing.T) {
 							},
 						},
 					},
+					{
+						From: &api.ImageObjRef{
+							Kind: "DockerImage",
+							ObjectReference: api.ObjectReference{
+								Name: "quay.io/custom/image:1.0",
+							},
+						},
+					},
 				},
 			},
 		},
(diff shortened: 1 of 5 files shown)


desmax74 (approved):

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -3432,7 +3461,7 @@ func TestMultipleBuildConfigurations(t *testing.T) {
 	env, err := GetEnvironment(cr, test.MockService())
 
 	assert.Nil(t, err, "Error getting prod environment")
-	assert.Len(t, env.Servers, 2, "Expect two KIE Servers to be created based on provided build configs")
+	assert.Len(t, env.Servers, 3, "Expect two KIE Servers to be created based on provided build configs")
```
jakubschwan, line 3464: Is this asser correct and should be there 3 server? If so it would be good to correctly update comment as well.

jakubschwan (commented):
added one comment, otherwise changes lgtm

pkg/controller/kieapp/defaults/defaults_test.go
spolti: it is correct, the build object I've added is a new server object.

PR #732, opened 2022-01-25
[KIECLOUD-570] Hide adminPassword from Business automation operator kieapp instance

desmax74:
See:https://issues.redhat.com/browse/KIECLOUD-570
Signed-off-by: desmax74 mdessi@redhat.com

### DIFF
--- a/deploy/crds/kieapp.crd.yaml
+++ b/deploy/crds/kieapp.crd.yaml
@@ -346,6 +346,11 @@ spec:
                     description: The password to use for keystore generation.
                     format: password
                     type: string
+                  secretAdminCredentials:
+                    description: Is the name of the secret containing admin's credentials,
+                      the keys are username and password, if the adminUser and adminPassword
+                      are provided the secretAdminCredentials will be skipped
+                    type: string
                   startupStrategy:
                     description: Startup strategy for Console and Kieserver
                     properties:
@@ -3199,6 +3204,12 @@ spec:
                       keyStorePassword:
                         description: The password to use for keystore generation.
                         format: password
+                        type: string
+                      secretAdminCredentials:
+                        description: Is the name of the secret containing admin's
+                          credentials, the keys are username and password, if the
+                          adminUser and adminPassword are provided the secretAdminCredentials
+                          will be skipped
                         type: string
                       startupStrategy:
                         description: Startup strategy for Console and Kieserver
--- a/deploy/ui/form.json
+++ b/deploy/ui/form.json
@@ -107,26 +107,64 @@
           ]
         },
         {
-          "label": "Admin user",
-          "type": "fieldGroup",
+          "label": "Admin user configuration",
+          "type": "dropDown",
           "required": false,
-          "jsonPath": "$.spec.commonConfig",
+          "jsonPath": "$.spec.commonConfig.AdminUser",
           "visible": true,
-          "fields": [
+          "default": "default",
+          "options": [
             {
-              "label": "Username",
-              "default": "adminUser",
-              "required": false,
-              "description": "Administrative username for RHPAM components",
-              "jsonPath": "$.spec.commonConfig.adminUser",
-              "type": "text"
+              "value": "secret",
+              "label": "Secret configuration"
             },
             {
-              "label": "Password",
+              "value": "default",
+              "label": "Username and Password configuration"
+            }
+          ],
+          "fields": [
+            {
+              "label": "Username and password to create a kie-admin-credentials secret",
+              "type": "fieldGroup",
               "required": false,
-              "description": "Administrative password to use instead of generating default value",
-              "jsonPath": "$.spec.commonConfig.adminPassword",
-              "type": "password"
+              "jsonPath": "$.spec.commonConfig",
+              "displayWhen": "default",
+              "visible": true,
+              "fields": [
+                {
+                  "label": "Username",
+                  "default": "adminUser",
+                  "required": false,
+                  "description": "Administrative username for RHPAM components",
+                  "jsonPath": "$.spec.commonConfig.adminUser",
+                  "type": "text"
+                },
+                {
+                  "label": "Password",
+                  "required": false,
+                  "description": "Administrative password to use instead of generating default value",
+                  "jsonPath": "$.spec.commonConfig.adminPassword",
+                  "type": "password"
+                 }
+              ]
+            },
+            {
+              "label": "Secret to use",
+              "type": "fieldGroup",
+              "required": false,
+              "jsonPath": "$.spec.commonConfig",
+              "displayWhen": "secret",
+              "visible": false,
+              "fields": [
+                {
+                  "label": "Secret configuration to use for Admin user",
+                  "required": false,
+                  "description": "If empty kie-admin-credentials secret will be used, if missing one with default username and generated password will be created",
+                  "jsonPath": "$.spec.commonConfig.secretAdminCredentials",
+                  "type": "text"
+                }
+              ]
             }
           ]
         },
--- a/pkg/apis/app/v2/kieapp_types.go
+++ b/pkg/apis/app/v2/kieapp_types.go
@@ -769,7 +769,8 @@ type ServerTemplate struct {
 	Cors                   *CORSFiltersObject            `json:"cors,omitempty"`
 	StartupStrategy        *StartupStrategy              `json:"startupStrategy,omitempty"`
 	// MDBMaxSession number of KIE Executor sessions
-	MDBMaxSession *int `json:"MDBMaxSession,omitempty"`
+	MDBMaxSession          *int   `json:"MDBMaxSession,omitempty"`
+	SecretAdminCredentials string `json:"secretAdminCredentials,omitempty"`
 }
 
 // DashbuilderTemplate contains all the variables used in the yaml templates
@@ -941,6 +942,8 @@ type CommonConfig struct {
 	DisableSsl bool `json:"disableSsl,omitempty"`
 	// Startup strategy for Console and Kieserver
 	StartupStrategy *StartupStrategy `json:"startupStrategy,omitempty"`
+	// Is the name of the secret containing admin's credentials, the keys are username and password, if the adminUser and adminPassword are provided the secretAdminCredentials will be skipped
+	SecretAdminCredentials string `json:"secretAdminCredentials,omitempty"`
 }
 
 // VersionConfigs ...
--- a/pkg/controller/kieapp/constants/constants.go
+++ b/pkg/controller/kieapp/constants/constants.go
@@ -292,6 +292,10 @@ const (
 	SUBCOMPONENT_TYPE_INFRA = "infrastructure"
 
 	DefaultDatagridUsername = "infinispan"
+
+	KIE_ADMIN_CREDENTIALS_SECRET = "kie-admin-credentials"
+	USERNAME_ADMIN_SECRET_KEY    = "username"
+	PASSWORD_ADMIN_SECRET_KEY    = "password"
 )
 
 var OseCli4ImageURL = ImageRegistry + "/openshift4/ose-cli:" + highestOcpVersion(Ocp4Versions)
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -5,6 +5,7 @@ package defaults
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -60,6 +61,29 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 		cr.Status.Applied.Version = constants.CurrentVersion
 		cr.Spec.Version = ""
 	}
+
+	// if the secret is the only credential present
+	if len(cr.Spec.CommonConfig.SecretAdminCredentials) > 0 && len(cr.Spec.CommonConfig.AdminUser) == 0 && len(cr.Spec.CommonConfig.AdminPassword) == 0 {
+		creds := constants.KIE_ADMIN_CREDENTIALS_SECRET
+		if len(cr.Spec.CommonConfig.SecretAdminCredentials) > 0 {
+			creds = cr.Spec.CommonConfig.SecretAdminCredentials
+		}
+		errSecret, _, _ := checkAndCreateAdminSecret(service, creds, cr.Namespace, cr.Spec.Environment)
+		if errSecret != nil {
+			log.Error("Can't create Admin Secret. ", errSecret)
+		}
+	}
+
+	//in case of missing secretAdmin we use the AdminUsername AdminPassword and if missing we set the defaults
+	if len(cr.Spec.CommonConfig.SecretAdminCredentials) == 0 && len(cr.Spec.CommonConfig.AdminPassword) == 0 {
+		password := constants.DefaultPassword
+
+		if isTrial(cr) {
+			password = string(shared.GeneratePassword(8))
+		}
+		cr.Spec.CommonConfig.AdminPassword = password
+	}
+
 	envTemplate, err := getEnvTemplate(cr)
 	if err != nil {
 		return api.Environment{}, err
@@ -761,17 +785,18 @@ func getServersConfig(cr *api.KieApp) ([]api.ServerTemplate, error) {
 			}
 			usedNames[name] = true
 			template := api.ServerTemplate{
-				KieName:          name,
-				KieServerID:      name,
-				Build:            getBuildConfig(product, cr, serverSet),
-				KeystoreSecret:   serverSet.KeystoreSecret,
-				StorageClassName: serverSet.StorageClassName,
-				JbpmCluster:      serverSet.JbpmCluster,
-				PersistRepos:     serverSet.PersistRepos,
-				ServersM2PvSize:  serverSet.ServersM2PvSize,
-				ServersKiePvSize: serverSet.ServersKiePvSize,
-				StartupStrategy:  cr.Status.Applied.CommonConfig.StartupStrategy,
-				MDBMaxSession:    serverSet.MDBMaxSession,
+				KieName:                name,
+				KieServerID:            name,
+				Build:                  getBuildConfig(product, cr, serverSet),
+				KeystoreSecret:         serverSet.KeystoreSecret,
+				StorageClassName:       serverSet.StorageClassName,
+				JbpmCluster:            serverSet.JbpmCluster,
+				PersistRepos:           serverSet.PersistRepos,
+				ServersM2PvSize:        serverSet.ServersM2PvSize,
+				ServersKiePvSize:       serverSet.ServersKiePvSize,
+				StartupStrategy:        cr.Status.Applied.CommonConfig.StartupStrategy,
+				MDBMaxSession:          serverSet.MDBMaxSession,
+				SecretAdminCredentials: cr.Status.Applied.CommonConfig.SecretAdminCredentials,
 			}
 
 			if cr.Status.Applied.Objects.Console == nil || cr.Status.Applied.Environment == api.RhdmProductionImmutable {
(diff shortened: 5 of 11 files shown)


desmax74:
@jakubschwan @spolti

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -2089,3 +2114,55 @@ func getRouteHostname(obj interface{}) (host string) {
...
 	return host
 }
+
+func checkAndCreateAdminSecret(service kubernetes.PlatformService, namespace string, secretName string) (error, string, string) {
+	customerSecret, err := getSecret(service, namespace, secretName)
+
+	// The secret passed by the user doesn't exist we create one with default values
+	if err != nil {
+		username := constants.DefaultAdminUser
+		password := string(shared.GeneratePassword(8))
```
desmax74, line 2124: @spolti is it ok use a generated password or we want to use the default password like in the trial env ?

tchughesiv (changes requested):

tchughesiv (commented):
how will operator upgrades be handled? existing KieApps using admin/pwd in yaml?

deploy/crds/kieapp.crd.yaml
```diff
@@ -262,7 +262,6 @@ spec:
                     type: string
                   dbPassword:
                     description: The password to use for databases.
-                    format: password
```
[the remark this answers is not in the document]
desmax74: Isn't a manual remove, let me check why is removed during the make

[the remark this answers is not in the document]
desmax74: Checked the text isn't in clear text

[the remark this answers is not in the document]
tchughesiv: please run a make and be sure it isn't removed... thanks!

spolti:
> how will operator upgrades be handled? existing KieApps using admin/pwd in yaml?

That's a good question, I think we have two options:
- keep both and mark these fields as deprecated somehow to be removed when we update the api.
- wait until 7.13 e do a api update.

deploy/crds/kieapp.crd.yaml
```diff
@@ -287,6 +299,8 @@ spec:
                           default is OpenShiftStartupStrategy
                         type: string
                     type: object
+                required:
+                - secretAdminCredentialsReference
```
jakubschwan, line 303: Not sure if we want to have it as a required parameter. Users may still prefer setting up the user using the adminUser and adminPassword string property (in commonConfig).
I would probably suggest to remove it from required.

deploy/crds/kieapp.crd.yaml
```diff
@@ -272,6 +272,18 @@ spec:
                     description: The password to use for keystore generation.
                     format: password
                     type: string
+                  secretAdminCredentialsReference:
+                    description: SecretAdminReference is a reference to the secret
+                      containing this user's password
+                    format: password
```
jakubschwan, line 278: is this format really needed for setting up string property of some secret name?

deploy/crs/v2/snippets/common_config.yaml
```diff
@@ -11,4 +11,5 @@ spec:
   commonConfig:
     # By default is the cr.Name
     applicationName: other
-    adminPassword: example
+    secretAdminCredentialsReference:
+      name: "kie-admin-credentials"
```
jakubschwan, line 15: let's create a new snippet with example of usage of user credentials secret

deploy/ui/form.json
```diff
@@ -107,26 +107,63 @@
...
-              "label": "Username",
-              "default": "adminUser",
-              "required": false,
-              "description": "Administrative username for RHPAM components",
-              "jsonPath": "$.spec.commonConfig.adminUser",
-              "type": "text"
+              "label": "Secret to use",
+              "type": "fieldGroup",
+              "required": true,
+              "jsonPath": "$.spec",
```
jakubschwan, line 130: Is in group needed json path, and if so there should be used the commonConfig object as well, right?
```suggestion
              "jsonPath": "$.spec.commonConfig",
```

deploy/ui/form.json
```diff
@@ -107,26 +107,63 @@
...
+              "displayWhen": "secret",
+              "visible": true,
+              "fields": [
+                {
+                  "label": "Secret configuration to use for Admin user",
+                  "type": "text",
+                  "jsonPath": "$.spec.commonConfig.adminSecret",
+                  "originalJsonPath": "$.spec.commonConfig",
+                  "description": "If empty kie-admin-credentials secret will be used, if missing one with default username and generated password will be created"
```
jakubschwan, line 139: Is default username and generated secret true? according the KieApp CRD the secret name is required parameter, so I would expect it needs to be set by user.

deploy/ui/form.json
```diff
@@ -107,26 +107,63 @@
...
+              "type": "fieldGroup",
+              "required": true,
+              "jsonPath": "$.spec",
+              "displayWhen": "secret",
+              "visible": true,
+              "fields": [
+                {
+                  "label": "Secret configuration to use for Admin user",
+                  "type": "text",
+                  "jsonPath": "$.spec.commonConfig.adminSecret",
```
jakubschwan, line 137: Wrongly used jsonPath, it should be something like this:
```suggestion
                  "jsonPath": "$.spec.commonConfig.secretAdminCredentialsReference.name",
```
please check all the UI form and be sure that generated KieApp from this form can be run by operator.

pkg/controller/kieapp/constants/constants.go
```diff
@@ -292,6 +292,10 @@ const (
 	SUBCOMPONENT_TYPE_INFRA = "infrastructure"
 
 	DefaultDatagridUsername = "infinispan"
+
+	KIE_ADMIN_CREDENTIALS_SECRET = "kie-admin-credentials"
+	USERNAME_ADMIN_SECRET_KEY    = "username"
+	PASSWORD_ADMIN_SECRET_KEY    = "password"
```
jakubschwan, line 298: The format of the credentials secret, or what keys are used, is not mentioned anywhere. So it would be good to add it to the desciption in KieApp.
Also it would be handy to add within this PR an example of credentials secret.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -60,6 +61,30 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
 		cr.Status.Applied.Version = constants.CurrentVersion
 		cr.Spec.Version = ""
 	}
+
+	/*  with cr.Spec.CommonConfig.SecretAdminCredentialsReference != nil && len(cr.Spec.CommonConfig.AdminUser) >= 0 && len(cr.Spec.CommonConfig.AdminPassword) >= 0
+	we do nothing because is a condition used in the tests and we skip the secret check (not available in the unit tests)
+	*/
```
jakubschwan, line 67: Do nothing is not ideal, inmo. If we run to this state, maybe it would be good to remove secretAdmin form status aplied section in KieApp CRD, so in that case users would see that this config was not applied.
Also it would be good to mentioned it in description that credentials secret will be not used if adminUser and adminPassword is already set.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -60,6 +61,30 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+	if cr.Spec.CommonConfig.SecretAdminCredentialsReference != nil && len(cr.Spec.CommonConfig.AdminUser) == 0 && len(cr.Spec.CommonConfig.AdminPassword) == 0 {
+		errSecret, adminUsername, adminPassword := checkAndCreateAdminSecret(service, cr.Spec.CommonConfig.SecretAdminCredentialsReference.Name, cr.Namespace, cr.Spec.Environment)
+		if errSecret != nil {
+			log.Error("Can't create Admin Secret. ", errSecret)
+			// here we continue instead of return otherwise fail tests without k8s runtime to see the created secret
+		}
+		cr.Spec.CommonConfig.AdminUser = adminUsername
+		cr.Spec.CommonConfig.AdminPassword = adminPassword
```
jakubschwan, line 75: As we want to hide these values, we should add them here to cr. This would be then displayed in KieApp, right? probably leftover from debugging.
```suggestion
```

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -60,6 +61,30 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+	}
+
+	if cr.Spec.CommonConfig.SecretAdminCredentialsReference == nil && len(cr.Spec.CommonConfig.AdminUser) == 0 && len(cr.Spec.CommonConfig.AdminPassword) == 0 {
+		errSecret, adminUsername, adminPassword := checkAndCreateAdminSecret(service, constants.KIE_ADMIN_CREDENTIALS_SECRET, cr.Namespace, cr.Spec.Environment)
+		if errSecret != nil {
+			log.Error("Can't create Admin Secret. ", errSecret)
+			// here we continue instead of return otherwise fails test without k8s runtime to see the created secret
+		}
+		cr.Spec.CommonConfig.AdminUser = adminUsername
+		cr.Spec.CommonConfig.AdminPassword = adminPassword
```
jakubschwan, line 85: As we want to hide these values, we should add them here to cr. This would be then displayed in KieApp, right? probably leftover from debugging.
```suggestion
```

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -2089,3 +2114,59 @@ func getRouteHostname(obj interface{}) (host string) {
 	}
 	return host
 }
+
+func checkAndCreateAdminSecret(service kubernetes.PlatformService, namespace string, secretName string, environment api.EnvironmentType) (error, string, string) {
+	customerSecret, err := getSecret(service, namespace, secretName)
+
+	// The secret passed by the user doesn't exist we create one with default values
+	if err != nil {
```
jakubschwan, line 2122: is this check correct? If secret exists, then is return err ??

jakubschwan (changes requested):

deploy/crds/kieapp.crd.yaml
desmax74: No

deploy/crds/kieapp.crd.yaml
desmax74: Ok

deploy/ui/form.json
desmax74: yep

pkg/controller/kieapp/defaults/defaults.go
desmax74: If secret exists the err is nil

deploy/ui/form.json
desmax74: Removed the required

pkg/controller/kieapp/constants/constants.go
desmax74: ok

deploy/crs/v2/snippets/common_config.yaml
desmax74: ok

deploy/ui/form.json
```diff
@@ -107,26 +107,63 @@
...
             {
-              "label": "Password",
-              "required": false,
-              "description": "Administrative password to use instead of generating default value",
-              "jsonPath": "$.spec.commonConfig.adminPassword",
-              "type": "password"
+              "label": "Username and password to create a kie-admin-credentials secret",
+              "type": "fieldGroup",
+              "required": true,
+              "jsonPath": "$.spec",
```
jakubschwan, line 147: For previous group, the Secret to use, is used little different jsonPath, shouldn't be this two fields same?
```suggestion
              "jsonPath": "$.spec.commonConfig",
```

deploy/ui/form.json
```diff
@@ -107,26 +107,63 @@
...
+                }
+              ]
             },
             {
-              "label": "Password",
-              "required": false,
-              "description": "Administrative password to use instead of generating default value",
-              "jsonPath": "$.spec.commonConfig.adminPassword",
-              "type": "password"
+              "label": "Username and password to create a kie-admin-credentials secret",
```
jakubschwan, line 144: Will be really credentials secret created? In this case this can be breaking change and we probably don't wont to do that. AdminUser and AdminPassword should be still creating the user in the old way, without the secret.
```suggestion
              "label": "Username and password to create",
```
Or is this just some typo or copy paste leftover?

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -894,6 +895,8 @@ type CommonConfig struct {
 	DisableSsl bool `json:"disableSsl,omitempty"`
 	// Startup strategy for Console and Kieserver
 	StartupStrategy *StartupStrategy `json:"startupStrategy,omitempty"`
+	// Is the name of the secret containing admin's credentials, the keys are username and password, if the username and password are provided the secretAdminCredentials will be skipped
```
jakubschwan, line 898: It is confusing when there used same names for keys and properties. Also properties are named differently so in that case it would be much better and cleaner to use the property names from KieApp.
```suggestion
	// Is the name of the secret containing admin's credentials, the keys are username and password, if the adminUser and adminPassword are provided the secretAdminCredentials will be skipped
```

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -60,6 +61,44 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+	}
+
+	// we create the kie admin secret with the provided username and password
+	if len(cr.Spec.CommonConfig.SecretAdminCredentials) == 0 && len(cr.Spec.CommonConfig.AdminUser) > 0 && len(cr.Spec.CommonConfig.AdminPassword) > 0 {
+		errSecret := createSecret(service, cr.Namespace, constants.KIE_ADMIN_CREDENTIALS_SECRET, cr.Spec.CommonConfig.AdminUser, cr.Spec.CommonConfig.AdminPassword)
+		if errSecret != nil {
+			log.Error("Can't create Admin Secret. ", errSecret)
+			// here we continue instead of return otherwise fails test without k8s runtime to see the created secret
+		}
+	}
```
jakubschwan, line 100: Not sure how exactly are all of these checks used and needed. Maybe it's little to complicated.
Secret should be used only in case that is set the secret name and there is no configuration for adminUser and adminPassword, right?
If there is no secret or any config for adminUser and adminPwd, then are used generated values for adminUser and adminPassword.
In that case wouldn't be easier to just check that `secret > 0 & adminUser = 0 & adminPwd = 0` ?

jakubschwan (approved):

deploy/ui/form.json
desmax74: This is the new behaviour decided with Felipe, no more username and password in clear text but in the secrets

deploy/ui/form.json
desmax74: fixed

deploy/ui/form.json
desmax74: Works with just "$.spec" mysteries  of the console-form-cr...

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -2100,3 +2123,74 @@ func getRouteHostname(obj interface{}) (host string) {
...
+	type: Opaque
+	data:
+	  //adminUser
+	  username: YWRtaW4=
+	  //RedHat
+	  password: UmVkSGF0
+
+	*/
+
+	customerSecret, err := getSecret(service, namespace, secretName)
```
spolti, line 2143: better to rename it, something like providedSecret

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -2100,3 +2123,74 @@ func getRouteHostname(obj interface{}) (host string) {
...
+
+	// The secret passed by the user doesn't exist we create one with default values
+	if err != nil {
+		username := constants.DefaultAdminUser
```
spolti, line 2147: probably it is a good idea remove the blank user and pasword from cr and cr.applied after the secret is in place.
@jakubschwan @desmax74 wdyt?

spolti (changes requested):

pkg/apis/app/v2/kieapp_types.go
```diff
@@ -769,7 +769,8 @@ type ServerTemplate struct {
 	Cors                   *CORSFiltersObject            `json:"cors,omitempty"`
 	StartupStrategy        *StartupStrategy              `json:"startupStrategy,omitempty"`
 	// MDBMaxSession number of KIE Executor sessions
-	MDBMaxSession *int `json:"MDBMaxSession,omitempty"`
+	MDBMaxSession          *int   `json:"MDBMaxSession,omitempty"`
+	SecretAdminCredentials string `json:"secretAdminCredentials,omitempty"`
```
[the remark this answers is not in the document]
desmax74: bad rebase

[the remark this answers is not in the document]
desmax74: Yes this line is for the template , on the commonconfig is here https://github.com/kiegroup/kie-cloud-operator/pull/732/files#diff-4d98202026f495310fe4aa8f64f71057eb9fa83104da7480469153b7d6fe810bR946

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -60,6 +61,28 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+		}
+		errSecret, _, _ := checkAndCreateAdminSecret(service, creds, cr.Namespace, cr.Spec.Environment)
+		if errSecret != nil {
+			log.Error("Can't create Admin Secret. ", errSecret)
+		}
+	}
+
+	//in case of missing secretAdmin we use the AdminUsername AdminPassword and if missing we set the defaults
+	if len(cr.Spec.CommonConfig.SecretAdminCredentials) == 0 && len(cr.Spec.CommonConfig.AdminPassword) == 0 {
+		password := constants.DefaultPassword
```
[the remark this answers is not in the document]
desmax74: The legacy behaviour was to generate just for the trial

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -5,6 +5,7 @@ package defaults
 import (
 	"bytes"
 	"context"
+	errors2 "errors"
```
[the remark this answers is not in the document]
desmax74: fixed

pkg/controller/kieapp/defaults/defaults.go
desmax74: Depends maybe someone is still using the "old" behaviour and is fine for they, now they could choose.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -60,6 +61,28 @@ func GetEnvironment(cr *api.KieApp, service kubernetes.PlatformService) (api.Env
...
+		errSecret, _, _ := checkAndCreateAdminSecret(service, creds, cr.Namespace, cr.Spec.Environment)
+		if errSecret != nil {
+			log.Error("Can't create Admin Secret. ", errSecret)
+		}
+	}
+
+	//in case of missing secretAdmin we use the AdminUsername AdminPassword and if missing we set the defaults
+	if len(cr.Spec.CommonConfig.SecretAdminCredentials) == 0 && len(cr.Spec.CommonConfig.AdminPassword) == 0 {
+		password := constants.DefaultPassword
+		if !strings.HasSuffix(string(cr.Spec.Environment), constants.TrialEnvSuffix) {
```
spolti, line 80: there is a isTrial func

desmax74: fixed

desmax74:
replaced by https://github.com/kiegroup/kie-cloud-operator/pull/758 on the main branch

PR #762, opened 2022-03-23
[RHPAM-4148] - LDAP authentication fails because direct-verification=…

spolti:
…true is not set in LDAP realm
See: https://issues.redhat.com/browse/RHPAM-4148


desmax74:
lgtm

PR #764, opened 2022-03-24
[RHPAM-4125] Add capability to configure Route termination

desmax74:


pkg/apis/app/v2/kieapp_types.go
```diff
@@ -154,18 +154,31 @@ type KieServerSet struct {
...
 	DecisionsOnly *bool `json:"decisionsOnly,omitempty"`
 	// Number of max KIE Executor sessions, it must be lower than the value of max-pool-size, by default is max-pool-size set to 60. Max pool size can be set by system property jboss.mdb.strict.max.pool.size (using javaOptsAppend: "-Djboss.mdb.strict.max.pool.size=40"), for more information see https://access.redhat.com/solutions/2955481.
-	KieExecutorMDBMaxSession *int `json:"kieExecutorMDBMaxSession,omitempty"`
+	KieExecutorMDBMaxSession *int              `json:"kieExecutorMDBMaxSession,omitempty"`
+	TerminationRoute         *TerminationRoute `json:"terminationRoute,omitempty"`
+}
+
+type TerminationRoute struct {
+	// Enable edge termination
+	EnableEdge bool `json:"enableEdge,omitempty"`
```
jakubschwan, line 163: @desmax74 you mentioned that this is only supported for http, right? We should mentioned it in description as well. And probably also that it is required to have set disableSsl to true as well.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -601,17 +613,16 @@ func getSpecEnv(envs []corev1.EnvVar, name string) string {
...
 
-func getSmartRouterProtocol(cr *api.KieApp) string {
-	if cr.Status.Applied.Objects.SmartRouter != nil {
-		if len(cr.Status.Applied.Objects.SmartRouter.Protocol) == 0 && cr.Status.Applied.CommonConfig.DisableSsl {
-			return constants.HttpProtocol
-		} else if len(cr.Status.Applied.Objects.SmartRouter.Protocol) == 0 && !cr.Status.Applied.CommonConfig.DisableSsl {
-			return constants.HttpsProtocol
-		} else {
-			return cr.Status.Applied.Objects.SmartRouter.Protocol
-		}
```
jakubschwan, line 612: This is dropped completly? If this property is still there we should probably keep it here.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1479,6 +1506,36 @@ func SetDefaults(cr *api.KieApp) {
...
+	if console.TerminationRoute == nil {
+		console.TerminationRoute = &api.TerminationRoute{}
+	}
+}
+
+func checkTerminationRouteOnSmartRouter(smartrouter *api.SmartRouterObject) {
+	if smartrouter.TerminationRoute == nil {
+		smartrouter.TerminationRoute = &api.TerminationRoute{}
+	}
+}
```
jakubschwan, line 1537: For what exactly are these checks needed? Are they configuring termination (adding termination property into kieapp) by default - if so this is probably something that we won't do in operator.

pkg/apis/app/v2/kieapp_types.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults_test.go
```diff
@@ -4979,6 +4982,7 @@ func TestGetProcessMigrationTemplate(t *testing.T) {
 									JavaMaxMemRatio:     Pint32(20),
 									JavaInitialMemRatio: Pint32(25),
 								},
+								TerminationRoute: &api.TerminationRoute{},
```
jakubschwan, line 4985: Why is this added to this tests? We should keep it as before.

jakubschwan (approved):
one minor thing, otherwise it looks good

pkg/controller/kieapp/defaults/defaults_test.go
desmax74: right, removed

spolti (approved):

PR #769, opened 2022-03-28
[KIECLOUD-632] - Update ocp version to 4.10 on ba-operator and operat…

spolti:
…or-utils


spolti:
needs to update the go version on CI.

desmax74 (approved):

PR #773, opened 2022-04-20
[KIECLOUD-636] - Prepare kie-cloud-operator for 7.13.0 and 7.13.1

spolti:
release

- remove previous config version


spolti:
this test is expected to fail since we are removing the previous version configuration files.

PR #779, opened 2022-04-26
[RHPAM-4140] Hide adminPassword from Business automation operator kieapp instance

desmax74:

### DIFF
--- a/pkg/controller/kieapp/defaults/defaults.go
+++ b/pkg/controller/kieapp/defaults/defaults.go
@@ -175,13 +175,6 @@ func handleAdminSecretUsernameAndPassword(cr *api.KieApp, service kubernetes.Pla
 		checkAndCreate(cr, cr.Spec.CommonConfig.AdminPassword, service)
 	} else if len(cr.Spec.CommonConfig.SecretAdminCredentials) > 0 && len(cr.Spec.CommonConfig.AdminPassword) == 0 {
 		checkAndCreate(cr, "", service)
-	} else if len(cr.Spec.CommonConfig.AdminPassword) == 0 {
-		//in case of missing secretAdmin we use the AdminUsername AdminPassword and if missing we set the defaults
-		password := constants.DefaultPassword
-		if !isTrial(cr) {
-			password = string(shared.GeneratePassword(8))
-		}
-		cr.Spec.CommonConfig.AdminPassword = password
 	}
 }
 
@@ -532,6 +525,8 @@ func getConsoleTemplate(cr *api.KieApp) api.ConsoleTemplate {
 		if cr.Status.Applied.Environment == api.RhpamAuthoringHA || cr.Status.Applied.Environment == api.RhdmAuthoringHA {
 			if cr.Status.Applied.Objects.Console.DataGridAuth != nil {
 				template.DataGridAuth = *cr.Status.Applied.Objects.Console.DataGridAuth
+			} else {
+				template.DataGridAuth = api.DataGridAuth{}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1266,13 +1259,31 @@ func getDefaultQueue(append bool, defaultJmsQueue string, jmsQueue string) strin
...
-func setPasswords(spec *api.KieAppSpec, isTrialEnv bool) {
+func setPasswords(spec *api.KieAppSpec, status api.KieAppStatus, isTrialEnv bool) {
 	passwords := []*string{
 		&spec.CommonConfig.KeyStorePassword,
 		&spec.CommonConfig.DBPassword,
 		&spec.CommonConfig.AMQPassword,
 		&spec.CommonConfig.AMQClusterPassword,
 	}
+
+	if len(spec.CommonConfig.SecretAdminCredentials) == 0 {
```
jakubschwan, line 1270: can you move this if to a new function. It would be better for understanding the code to read the function name then understanding what this if block does.

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1266,13 +1259,31 @@ func getDefaultQueue(append bool, defaultJmsQueue string, jmsQueue string) strin
...
+		} else if len(spec.CommonConfig.AdminPassword) == 0 {
+			spec.CommonConfig.AdminPassword = status.Applied.CommonConfig.AdminPassword
+		}
+	}
+
+	if spec.Objects.Console != nil && spec.Objects.Console.DataGridAuth != nil && len(spec.Objects.Console.DataGridAuth.Password) == 0 {
+		if status.Applied.Objects.Console != nil && status.Applied.Objects.Console.DataGridAuth != nil && len(status.Applied.Objects.Console.DataGridAuth.Password) > 0 {
+			if len(spec.Objects.Console.DataGridAuth.Password) == 0 {
+				spec.Objects.Console.DataGridAuth.Password = status.Applied.Objects.Console.DataGridAuth.Password
+			}
```
jakubschwan, line 1282: This block will be never executed as in previous if is check `len(status.Applied.Objects.Console.DataGridAuth.Password) > 0`

pkg/controller/kieapp/defaults/defaults.go
desmax74: ok

pkg/controller/kieapp/defaults/defaults.go
desmax74: Nope, one check is on spec and another is on status, but I can remove the  len(spec.Objects.Console.DataGridAuth.Password) == 0 
because is checked in the first if

davidesalerno (approved):

pkg/controller/kieapp/defaults/defaults.go
```diff
@@ -1285,6 +1285,28 @@ func setPasswords(spec *api.KieAppSpec, isTrialEnv bool) {
 	}
 }
 
+func handleDatagridAuth(spec *api.KieAppSpec, status api.KieAppStatus, passwords []*string) []*string {
+	if spec.Objects.Console != nil && spec.Objects.Console.DataGridAuth != nil && len(spec.Objects.Console.DataGridAuth.Password) == 0 {
+		if status.Applied.Objects.Console != nil && status.Applied.Objects.Console.DataGridAuth != nil && len(status.Applied.Objects.Console.DataGridAuth.Password) > 0 {
+			spec.Objects.Console.DataGridAuth.Password = status.Applied.Objects.Console.DataGridAuth.Password
```
jakubschwan, line 1291: shouldn't be also here set added password to passwords? Or it is already handled elsewhere?

desmax74: Here we are appending to the array of passwords if needed, depends on the information in the spec/status then are checked here https://github.com/kiegroup/kie-cloud-operator/pull/779/files/86e3d039a521ac7762be044765803d1e1453c21a#diff-29ad51b906bc49dd54db094a3b626e8e4a04628d03562e6c0d45db0e391ac2fbR1276

desmax74: Nope, we add to the array of passwords if a password must be included in the deploy and is empty. On Datagrid we check if the object are nil or not, instead on the kieadmin (mandatory for every deploy) we add to generate only if is empty, otherwise we set what is already deployed.

PR #834, opened 2022-10-25
[RHPAM-4464]-PostgreSQL 13 Pod won't start due to version 10 data directory

desmax74:
See:https://issues.redhat.com/browse/RHPAM-4464

### DIFF
--- a/pkg/controller/kieapp/constants/constants.go
+++ b/pkg/controller/kieapp/constants/constants.go
@@ -35,8 +35,8 @@ var VersionConstants = map[string]*api.VersionConfigs{
 		DatagridComponent:   Datagrid8Component,
 		MySQLImageURL:       MySQL80ImageURL,
 		MySQLComponent:      MySQL80Component,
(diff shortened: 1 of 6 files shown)


pkg/controller/kieapp/constants/constants.go
```diff
@@ -35,8 +35,8 @@ var VersionConstants = map[string]*api.VersionConfigs{
 		DatagridComponent:   Datagrid8Component,
 		MySQLImageURL:       MySQL80ImageURL,
 		MySQLComponent:      MySQL80Component,
-		PostgreSQLImageURL:  PostgreSQL12ImageURL,
-		PostgreSQLComponent: PostgreSQL12Component,
```
spolti, line 39: shouldn't we move this to the prior version or use PSSQL 13 in the prior version as well?

desmax74: ok

PR #841, opened 2022-12-04
[KIECLOUD-656] - Prepare BAMOE branches for 8.0.2

spolti:


spolti:
the config check is expected to fail since we are updating to 8.0.2 and removing old 8.0.0 config files.

desmax74 (approved):

PR #843, opened 2023-01-20
[KIECLOUD-659] Added first version of build-osbs.sh script for nightl…

davidesalerno:
…y build of operator and bundle

Signed-off-by: Davide Salerno <dsalerno@redhat.com>

See: [KIECLOUD-659](https://issues.redhat.com/browse/KIECLOUD-659)

Description:
The build-osbs.sh script (added into the hack folder with this change) will used by the Jenkins pipelines created by the new seed job that results from this [PR](https://github.com/kiegroup/kie-ci/pull/1407)

### DIFF
--- a/hack/build-osbs.sh
+++ b/hack/build-osbs.sh
@@ -1,0 +1,193 @@
+#!/bin/bash
+
+set -e
+
+DEBUG=
+GIT_USER=${GIT_USER:-"Your Name"}
+GIT_EMAIL=${GIT_EMAIL:-"yourname@email.com"}
+WORK_DIR=$(pwd)/build-temp
+CEKIT_RESPOND_YES=true
+
+function help()
+{
+    echo "usage: build-osbs.sh [options]"
+    echo
+    echo "Run a cekit osbs build of an rhpam ba operator image with bundle"
+    echo
+    echo "For each of the options below, the names of the arguments are environment variables that may be set"
+    echo "instead of using the particular option on the invocation"
+    echo ""
+    echo "Required:"
+    echo "  -v PROD_VERSION           Version being built. Passed to the build-overrides.sh -v option"
+    echo "  -t OSBS_BUILD_TARGET      Build target for osbs, for example rhba-7.3-openshift-containers-candidate"
+    echo ""
+    echo "Optional:"
+    echo "  -h                        Print this help message"
+    echo "  -p KERBEROS_PRINCIPAL     Kerberos principal to use with to access build systems. If not specified,"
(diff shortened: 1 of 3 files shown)


spolti (approved):

spolti:
@davidesalerno , isn't better to have this script centralized on another place, as it will be used by kogito as well?
maybe kie-ci repo?
 @rgdoliveira wdyt?

davidesalerno:
> @davidesalerno , isn't better to have this script centralized on another place, as it will be used by kogito as well? maybe kie-ci repo? @rgdoliveira wdyt?

Is there a way to put a bash script inside the kie-ci package and having it available inside the pipelines created by the seed job? Otherwise we have to download this script using wget or something similar from the Jenkins pipelines

lampajr (approved):

rgdoliveira:
> > @davidesalerno , isn't better to have this script centralized on another place, as it will be used by kogito as well? maybe kie-ci repo? @rgdoliveira wdyt?
> 
> Is there a way to put a bash script inside the kie-ci package and having it available inside the pipelines created by the seed job? Otherwise we have to download this script using wget or something similar from the Jenkins pipelines

Hi @davidesalerno @spolti (cc @winklerm),

I'm not sure you can do it in the way you suggested @davidesalerno . But another possible way would be to follow the same approach you used for the `download.sh` script (in gitlab eng-jenkins-scripts repository). Where you add the script and copy it to the Jenkins slaves and it is available in the system path. The con about this approach is that any change in the script would require the images to de recreated again.

Or in worst case maybe add it to https://github.com/kiegroup/kie-ci/tree/main/job-dsls/jobs/osbs and yeah, I suppose you will have to download it during the job run.

lampajr:
> I'm not sure you can do it in the way you suggested @davidesalerno 

Maybe he can, I'm thinking about using the same approach used here https://github.com/kiegroup/kie-ci/blob/main/job-dsls/jobs/kie/branch/kie_jenkinsScripts_PR.groovy#L22 and here https://github.com/kiegroup/kie-ci/blob/main/job-dsls/jobs/kie/branch/kie_jenkinsScripts_PR.groovy#L192, instead of defining the bash script inside the groovy script you can parse it with `def scriptTemplate = this.getClass().getResource("job-scripts/whatever-the-name.sh").text` and run it inside the pipeline using `sh <name of the variable>`. I never tried this approach though, but maybe it is worth to try at least one time. wdyt @rgdoliveira @davidesalerno ?

desmax74 (approved):

spolti:
Are you guys ok with the changes? can we merge it?

PR #847, opened 2023-02-16
[KIECLOUD-692] - Update the target dist-git branch with the new ones

spolti:


hack/image-osbs.patch
```diff
@@ -4,8 +4,8 @@ Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
 <+>UTF-8
```
rgdoliveira, line 4: as you are not changing the image.yaml file, you dont need to update this patch, otherwise I suppose git will fail to apply

davidesalerno (approved):

davidesalerno:
lgtm

hack/image-osbs.patch
rgdoliveira: never mind, ignore this comment.

rgdoliveira (approved):

desmax74 (approved):

PR #850, opened 2023-02-21
[KIECLOUD][RHPAM-4609] Setting up Cachito tool.

bacciotti:


image-prod.yaml
```diff
@@ -0,0 +1,55 @@
...
+        path: modules/builder-prod
+    install:
+      - name: builder
+
+  osbs:
+    configuration:
+      container:
+        image_build_method: imagebuilder
+        remote_source:
+          repo: https://github.com/kiegroup/kie-cloud-operator
```
spolti, line 18: leftover?

image-prod.yaml
```diff
@@ -0,0 +1,55 @@
+- schema_version: 1
```
spolti, line 1: is the builder a leftover?
It can probably be removed as it seems not be used.

modules/builder-prod/install
```diff
@@ -0,0 +1,23 @@
...
+
+OPERATOR_PATH=/usr/local/bin/kie-cloud-operator
+UI_PATH=/usr/local/bin/console-cr-form
+
+cd $REMOTE_SOURCE_DIR/app && pwd
+
+go generate ./...
+go vet ./...
+go test ./...
+CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -mod=vendor -a -o ${OPERATOR_PATH} ./cmd/manager
```
spolti, line 14: @rgdoliveira do you remember if the operator is build for power pc as well?

image-prod.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: "builder"
+  description: "Golang builder image"
+  version: ""
+  from: "registry.redhat.io/ubi8/go-toolset:1.17.7"
+  modules:
+    repositories:
```
spolti, line 7: you could use something similar:

```
  modules:
    repositories:
      - path: modules
    install:
      - name: rhpam.builder.prod
```

image-prod.yaml
spolti: nevermind.

image-prod.yaml
spolti, line 18: leftover?

image-prod.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - name: io.k8s.description
+      value: Operator for deploying RHPAM & RHDM
+    - name: io.k8s.display-name
+      value: Red Hat Business Automation Operator
+    - name: io.openshift.tags
+      value: rhpam,rhdm,operator
+  modules:
+    repositories:
+      - name: runtime
+        path: modules/runtime
```
spolti, line 45: where is this module?

I would expect to see something similar to:

```
  artifacts:
    - path: <path>
      image:builder
      dest: <dest>
```

modules/builder-prod/install
rgdoliveira: @spolti the RHPAM operator is built only for x86_64 - so the PR is correct. The operator that is built also for ppc64le is the Kogito operator.

image-prod.yaml
bacciotti: @spolti in this case the _install_ and _module.yaml_ files will be on modules folder root? Why don't keep it in modules/builder-prod path?

image-prod.yaml
bacciotti: @spolti are you talking about _repo_ and _ref_? They are used by Cachito.

image-prod.yaml
bacciotti: @spolti IIUC the artifacts are specified on _modules/runtime/module.yaml_ file.

image-prod.yaml
spolti: no, see the `path: modules` property.

image-prod.yaml
spolti: ok, but from where this runtime module is coming from?

image-prod.yaml
bacciotti: @spolti ok! changed to the suggested.

image-prod.yaml
bacciotti: Not sure, I got this file as example from image.yaml.

image-prod.yaml
spolti: Yes.
It is pointing to a specific ref.

how that will be handled and how the right branch will be selected?

will this value be changed on every build?

image-prod.yaml
spolti: which image.yaml?

spolti: I'd be surprised if the build don't fail, since CeKit will try to look for an non existent module.

image-prod.yaml
rgdoliveira: @spolti This is the correct approach FMPOV. When Prod team builds this image, we update the ref with the commit on the top of the branch, same as it is done for Kogito Operator, see https://github.com/kiegroup/kogito-operator/blob/1.13.x/rhpam-image-prod.yaml#L16-L28

spolti: > When Prod team builds this image, we update the ref with the commit on the top of the branch

ok, that answers my question :)

image-prod.yaml
rgdoliveira: @spolti isn't that modules/runtimes this one https://github.com/kiegroup/kie-cloud-operator/tree/release-v7.13.x/modules/runtime ? Sorry if I'm missing something.

Actually we had this same block in the original `image.yaml` file, see https://github.com/kiegroup/kie-cloud-operator/blob/release-v7.13.x/image.yaml#L36-L39, and we just based this new yaml on it.

spolti: you guys are correct, sry, my bad :)

spolti:
@bacciotti before merging can you please re-target the PR to the main branch?

7.13.x still pointing to 7.13.2 while main branch is for 7.13.3.

spolti:
we would need the same changes for the release-v7.13.x-blue branch.

bacciotti:
> @bacciotti before merging can you please re-target the PR to the main branch?
> 
> 7.13.x still pointing to 7.13.2 while main branch is for 7.13.3.

@spolti cc @rgdoliveira target changed to main.

rgdoliveira:
> > @bacciotti before merging can you please re-target the PR to the main branch?
> > 7.13.x still pointing to 7.13.2 while main branch is for 7.13.3.
> 
> @spolti cc @rgdoliveira target changed to main.

It is getting a lot of PRs to main branch too...I suppose you will have to do it by hand and not just changing the target?

bacciotti:
> > > @bacciotti before merging can you please re-target the PR to the main branch?
> > > 7.13.x still pointing to 7.13.2 while main branch is for 7.13.3.
> > 
> > 
> > @spolti cc @rgdoliveira target changed to main.
> 
> It is getting a lot of other commits to main branch too...I suppose you will have to do it by hand and not just changing the target?

wdyt @spolti ? I can recreate it from the scratch using main.

spolti:
I think you could cherry-pick this PR to main.

rgdoliveira:
Closed in favor of https://github.com/kiegroup/kie-cloud-operator/pull/853

PR #854, opened 2023-02-28
[release-v7.13.x-blue][KIECLOUD][RHPAM-4609] Setting up Cachito tool.

bacciotti:
**Related Pull Requests**
* https://github.com/kiegroup/kie-cloud-operator/pull/853
* https://github.com/kiegroup/kie-cloud-operator/pull/854

### DIFF
--- a/image-prod.yaml
+++ b/image-prod.yaml
@@ -7,7 +7,7 @@
     repositories:
       - path: modules/builder-prod
     install:
-      - name: rhpam.builder.prod
+      - name: bamoe.builder.prod
 
   osbs:
     configuration:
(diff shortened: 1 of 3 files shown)


image-prod.yaml
```diff
@@ -0,0 +1,54 @@
+- schema_version: 1
```
rgdoliveira, line 1: You will need to update most of the entries in this file to match the ones from [BAMOE image.yaml](https://github.com/kiegroup/kie-cloud-operator/blob/release-v7.13.x-blue/image.yaml). Please check the file I pointed out and update this one based on it.

modules/builder-prod/module.yaml
```diff
@@ -0,0 +1,6 @@
+schema_version: 1
+
+name: rhpam.builder.prod
```
rgdoliveira, line 3: rename this module name to bamoe.builder.prod?

rgdoliveira (changes requested):

image-prod.yaml
bacciotti: Good point. Changed!

PR #861, opened 2023-03-08
[KIECLOUD-699] - Override the commit reference in the rhpam/bamoe ope…

spolti:
…rator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ rhel-release:
 
 .PHONY: rhel-nightly
 rhel-nightly:
-	./hack/go-build.sh ${BUILDER} rhel nightly
+	./hack/go-build.sh ${BUILDER} rhel nightly $(build_user)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -84,7 +84,7 @@ rhel-release:
 
 .PHONY: rhel-nightly
 rhel-nightly:
-	./hack/go-build.sh ${BUILDER} rhel nightly
+	./hack/go-build.sh ${BUILDER} rhel nightly $(build_user)
```
davidesalerno, line 87: I've got a doubt, is it correct to use ```$(build_user)``` instead of ```${build_user}```

spolti: $() expect it as parameter from make , e.g.

make test build_user=user1

While ${} is a env.

davidesalerno (approved):
/lgtm

PR #984, opened 2025-01-17
Revert "Prepare BAMOE 8.0.6-5 bundle respin"

akumar074:
Reverts kiegroup/kie-cloud-operator#981


akumar074:
As we did not release the 8.0.6-4 bundle configuration, the 8.0.6-5 should not be added. We should keep the 8.0.6-4 config and rebuild the operator and bundle.

RishiRajAnand (approved):

Issue #987, opened 2025-02-05
Operator UI does not support custom hostname


sveno1990:
Currently the operator UI does not support custom hostnames, while the operator resources do. The option to specifiy a custom hostname for the operator ui using environment variables would be quite helpful in situations in which you the default hostname is different from the desired hostname . Is it possible to add the option?

PR #988, opened 2025-02-05
Fix custom hostname

sveno1990:
Attempt to fix https://github.com/kiegroup/kie-cloud-operator/issues/987.

### DIFF
--- a/pkg/controller/kieapp/deploy_ui.go
+++ b/pkg/controller/kieapp/deploy_ui.go
@@ -338,7 +338,7 @@ func getRoute(namespace string) *routev1.Route {
 			Labels:    labels,
 		},
 		Spec: routev1.RouteSpec{
-			Host: operatorUIHostname
+			Host: operatorUIHostname,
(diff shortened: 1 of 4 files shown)


sveno1990:
@jakubschwan or @rgdoliveira could one of you have a look?

rgdoliveira:
I will let @jakubschwan test this PR as he is from QE and has a better understanding if this can break something.

pkg/controller/kieapp/deploy_ui.go
```diff
@@ -337,6 +338,7 @@ func getRoute(namespace string) *routev1.Route {
 			Labels:    labels,
 		},
 		Spec: routev1.RouteSpec{
+			Host: operatorUIHostname
```
davidesalerno: What will it happen if the OPERATOR_UI_HOSTNAME is blank?

sveno1990: The route.spec.host field should accept "" as input.

PR #991, opened 2025-03-03
Fix replaces by in 8.0.7-1

RishiRajAnand:


deploy/olm-catalog/dev/8.0.7-1/manifests/bamoe-businessautomation-operator.clusterserviceversion.yaml
```diff
@@ -420,7 +420,7 @@ spec:
   maturity: dev
   provider:
     name: IBM
-  replaces: bamoe-businessautomation-operator.8.0.7-4
```
rgdoliveira, line 423: not only this file but other files need to change to correct version, please run the process again for updating it and make sure all places are updated.

rgdoliveira:
you cannot have it as 8.0.7-2, it needs to be 8.0.7-1 and have the replaces as bamoe-businessautomation-operator.8.0.6-4 in all the places it has a replace

rgdoliveira:
I think it looks good now

PR #1020, opened 2025-08-26
[ Draft ] Update RHPAM bundle for CVE respin

saumya1singh:


jakubschwan (commented):
Changes looks good, right now they are not required, as last available bundle has csv version ending with suffix -15

This PR can be labeled as do not merge yet until next rebuild.

PR #1048, opened 2026-05-27
Revert bundle updates to version 24 and 25

jakubschwan:
Latest available version of bundle is 22. The build of operator bundle image with version 24 or 25 will fail on checks, as there are non-existing versions to replace. This PR removes 2 latest version updates to fix the bundle image post build checks.


mareknovotny:
ok, i get this, @saumya1singh seems just took the latest even they were added to unreleased erratas.

mareknovotny (approved):
/lgtm

